PR #16, opened 2024-08-11
add firebase credentials and update deprecated code

alex-andria:
add firebase credentials and update deprecated code


tannaurus:
Thank you @alex-andria !

PR #19, opened 2024-08-12
#3 Updating app nav bar

### LINKED ISSUE #3: 3. As a user, I want to be able to navigate to all the pages in the application by clicking the links in the nav bar
## Summary

Users need to be able to navigate between the pages of our application in order to access all of its features. There is a navigation bar in the `Layout.jsx` component, but it's not quite feature-complete: the links in the nav element don't actually allow the user to navigate from page to page.

## Acceptance criteria
- [x] The `nav` element is added to the `Layout` component
- [x] Links to the "Home", "List", and "Add item" pages of the app are In the `nav` element
- [x] The links all function as expected using  the `NavLink` component from `react-router-dom`

## Notes
- The [official docs for the `NavLink` component](https://reactrouter.com/en/6.14.2/components/nav-link) will come in very handy!
- While a standard anchor element (`<a>`) _could_ allow a user to navigate from page to page, the `NavLink` component adds additional behavior that is essential to our app functioning correctly!

bbland1:

## Description

Updated nav bar element to use NavLink components in place of button elements to add page navigation.

## Related Issue

Closes #3 

## Acceptance Criteria

- [x] The `nav` element is added to the `Layout` component
- [x] Links to the "Home", "List", and "Add item" pages of the app are In the `nav` element
- [x] The links all function as expected using  the `NavLink` component from `react-router-dom`

## Type of Changes

`bug fix` `enhancement`

* From your terminal, pull down this branch with `git pull origin bb-ma/update-nav-bar-navigation` and check that branch out with `git checkout bb-ma/update-nav-bar-navigation` 
* Then `npm ci` to install the newly added dependencies locally and `npm start` to launch the app.
* Click the various navigations links to see new pages and verify functionality.

### DIFF
--- a/src/views/Layout.jsx
+++ b/src/views/Layout.jsx
@@ -28,24 +28,24 @@ export function Layout() {
 						{!!user ? <SignOutButton /> : <SignInButton />}
 						<NavLink
 							to="/"
-							className={({ isActive, isPending, isTransitioning }) =>
-								`Nav-link ${isPending ? 'pending' : ''} ${isActive ? 'active' : ''} ${isTransitioning ? 'transitioning' : ''}`
+							className={({ isActive }) =>
+								`Nav-link ${isActive ? 'active' : ''}`
 							}
 						>
 							Home
 						</NavLink>
 						<NavLink
 							to="/list"
-							className={({ isActive, isPending, isTransitioning }) =>
-								`Nav-link ${isPending ? 'pending' : ''} ${isActive ? 'active' : ''} ${isTransitioning ? 'transitioning' : ''}`
+							className={({ isActive }) =>
+								`Nav-link ${isActive ? 'active' : ''}`
 							}
 						>
 							List
 						</NavLink>
 						<NavLink
 							to="/manage-list"
-							className={({ isActive, isPending, isTransitioning }) =>
-								`Nav-link ${isPending ? 'pending' : ''} ${isActive ? 'active' : ''} ${isTransitioning ? 'transitioning' : ''}`
+							className={({ isActive }) =>
+								`Nav-link ${isActive ? 'active' : ''}`
 							}
 						>
 							Manage List
--- a/src/views/Layout.jsx
+++ b/src/views/Layout.jsx
@@ -26,13 +26,17 @@ export function Layout() {
 				<nav className="Nav">
 					<div className="Nav-container">
 						{!!user ? <SignOutButton /> : <SignInButton />}
-						<NavLink to="/" className="Nav-link">
+						<NavLink to="/" className="Nav-link" aria-label="Home">
 							Home
 						</NavLink>
-						<NavLink to="/list" className="Nav-link">
+						<NavLink to="/list" className="Nav-link" aria-label="List">
 							List
 						</NavLink>
-						<NavLink to="/manage-list" className="Nav-link">
+						<NavLink
+							to="/manage-list"
+							className="Nav-link"
+							aria-label="Manage List"
+						>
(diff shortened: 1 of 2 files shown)


bbland1 (commented):
@eternalmaha to make the draft PR I had to have a change to initiate it. The signin button that tanner set up just wasn't in the same container as the other buttons so I moved that in to make the draft PR we can definitely move it back or other changes for it but just wanted to let you know if you notice this one random change.

src/views/Layout.jsx
```diff
@@ -24,8 +24,8 @@ export function Layout() {
 					<Outlet />
 				</main>
 				<nav className="Nav">
-					{!!user ? <SignOutButton /> : <SignInButton />}
 					<div className="Nav-container">
+						{!!user ? <SignOutButton /> : <SignInButton />}
 						<button href="#" className="Nav-link">
 							Home
 						</button>
```
bbland1, line 31: So I know we will need to change these buttons to the NavLink comp from react-router-dom and I just not sure how the styling will work. 

I'm pretty sure it will be something like this but I know button elements have default styling so we may have to adjust the css once we make make sure the changing of the element works.

src/views/Layout.jsx
```diff
@@ -24,17 +24,32 @@ export function Layout() {
 					<Outlet />
 				</main>
 				<nav className="Nav">
-					{!!user ? <SignOutButton /> : <SignInButton />}
 					<div className="Nav-container">
-						<button href="#" className="Nav-link">
+						{!!user ? <SignOutButton /> : <SignInButton />}
+						<NavLink
+							to="/"
```
adidalal, line 30: Totally valid, however this can be as simple as something like `<NavLink to="/" className="Nav-link">Home</NavLink>`
Was there a specific reason to use active, pending, transitioning?

bbland1: Honestly the pending and transitioning not really 😅. I was looking at all those functionailties while making sure the NavLink works and I added them.

There was an active class to be set to the buttons when they were active originally, it bolds the underline of the word when that is the acttive class, so to keep that functionality working as I beleive was the goal the active would need to be there.

bbland1: But i did just remember i think active is added by default with NavLink so if we don't have the pending and transition we may not need it specifically set.

src/views/Layout.jsx
```diff
@@ -24,17 +24,17 @@ export function Layout() {
 					<Outlet />
 				</main>
 				<nav className="Nav">
-					{!!user ? <SignOutButton /> : <SignInButton />}
 					<div className="Nav-container">
-						<button href="#" className="Nav-link">
+						{!!user ? <SignOutButton /> : <SignInButton />}
+						<NavLink to="/" className="Nav-link">
```
zahrafalak, line 29: This looks great! One small suggestion here would be to consider adding ARIA attributes to the interactive navigation links to improve accessibility. For instance:
```
<NavLink to="/" className="Nav-link" aria-label="Home">
  Home
</NavLink>
```

bbland1: Good point! Definetly can add all the aria-label attributes to the links. Any others that you think should be added?

RossaMania:
Nice! I like how simple and readable this all is!

src/views/Layout.jsx
RossaMania: Today I learned about Aria labels! A couple weeks ago, I was just thinking "When I use React with React Router, I wonder what other accessibility features are available for screen readers????"

PR #20, opened 2024-08-14
Issue #2: Render the List and list items a user has access to

zahrafalak:

## Description

The most important feature of our app is the ability to look at their shopping lists to see what's on them, so the user knows what they need to buy.

## Acceptance Criteria

Acceptance criteria

- [x]  The useShoppingLists hook is used in App.jsx to get the shopping lists a user has access to from the database
- [x]  The lists are passed into the Home component as a prop named data
- [x]  The setListPath function is passed into the Home component so we can update the current list
- [x]  The useShoppingListData hook is used in App.jsx to get the items in the current list from the Firestore database
- [x]  The list is passed into the List component as a prop named data
- [x]  In Home.jsx, the SingleList component is used to render the name of each shopping list a user is subscribed to
- [x]  In List.jsx, the ListItem component is used to render the name of each item

## Type of Changes

Rendered List name and items a user has access to. Feature. 

### After
Screenshot 2024-08-14 at 10 50 35 AM
Screenshot 2024-08-14 at 10 51 05 AM

### DIFF
--- a/src/views/Home.jsx
+++ b/src/views/Home.jsx
@@ -2,23 +2,22 @@ import './Home.css';
 import { SingleList } from '../components';
 
 export function Home({ data, setListPath }) {
-	const dataPresent = data.length !== 0;
+	const hasList = data.length !== 0;
 	return (
 		<div className="Home">
 			<p>
 				Hello from the home (<code>/</code>) page!
 			</p>
 			<ul>
-				{dataPresent
-					? data.map((list, index) => (
-							<SingleList
-								key={index}
-								name={list.name}
-								path={list.path}
-								setListPath={setListPath}
-							/>
-						))
-					: null}
+				{hasList &&
+					data.map((list, index) => (
+						<SingleList
+							key={index}
+							name={list.name}
+							path={list.path}
+							setListPath={setListPath}
+						/>
+					))}
 			</ul>
 		</div>
 	);
--- a/src/views/Home.jsx
+++ b/src/views/Home.jsx
@@ -1,16 +1,24 @@
 import './Home.css';
+import { SingleList } from '../components';
 
 export function Home({ data, setListPath }) {
+	const dataPresent = data.length !== 0;
 	return (
 		<div className="Home">
 			<p>
 				Hello from the home (<code>/</code>) page!
 			</p>
 			<ul>
-				{/**
-				 * TODO: write some JavaScript that renders the `lists` array
-				 * so we can see which lists the user has access to.
-				 */}
+				{dataPresent
+					? data.map((list, index) => (
+							<SingleList
+								key={index}
+								name={list.name}
+								path={list.path}
+								setListPath={setListPath}
+							/>
+						))
+					: null}
 			</ul>
 		</div>
 	);
--- a/src/views/List.jsx
+++ b/src/views/List.jsx
@@ -1,16 +1,15 @@
 import { ListItem } from '../components';
 
 export function List({ data }) {
-	const dataPresent = data.length !== 0;
+	const hasItem = data.length !== 0;
 	return (
 		<>
 			<p>
 				Hello from the <code>/list</code> page!
 			</p>
 			<ul>
-				{dataPresent
-					? data.map((item) => <ListItem key={item.id} name={item.name} />)
-					: null}
+				{hasItem &&
+					data.map((item) => <ListItem key={item.id} name={item.name} />)}
 			</ul>
 		</>
 	);
--- a/src/views/List.jsx
+++ b/src/views/List.jsx
@@ -1,17 +1,16 @@
 import { ListItem } from '../components';
 
 export function List({ data }) {
+	const dataPresent = data.length !== 0;
 	return (
 		<>
 			<p>
 				Hello from the <code>/list</code> page!
 			</p>
 			<ul>
-				{/**
-				 * TODO: write some JavaScript that renders the `data` array
-				 * using the `ListItem` component that's imported at the top
-				 * of this file.
-				 */}
+				{dataPresent
+					? data.map((item) => <ListItem key={item.id} name={item.name} />)
+					: null}
 			</ul>
 		</>
 	);
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -34,6 +34,7 @@ export function App() {
 	 * Check ./api/firestore.js for its implementation.
 	 */
 	const lists = useShoppingLists(userId, userEmail);
+
 	/**
 	 * This custom hook takes our token and fetches the data for our list.
 	 * Check ./api/firestore.js for its implementation.


src/views/Home.jsx
```diff
@@ -1,16 +1,24 @@
...
 		<div className="Home">
 			<p>
 				Hello from the home (<code>/</code>) page!
 			</p>
 			<ul>
-				{/**
-				 * TODO: write some JavaScript that renders the `lists` array
-				 * so we can see which lists the user has access to.
-				 */}
+				{dataPresent
```
bbland1, line 12: It's awesome to actually see a list once I added the sharedList field to myself!

Was there a reason y'all chose to create a specific variable for the dataPresent over possibly doing it inline I guess with the mapping of the data? 

This could totally be a nitpick so as Alex put in slack I'm gonna put `NIT` haha because this is totally valid just curious.

```suggestion
{data.length !== 0 ? data.map((list, index) => {<SingleList key={index} name={list.name} path={list.path} setListPath={setListPath} /> : null}
```

bbland1, line 12: It's awesome to actually see a list once I added the sharedList field to myself!

Was there a reason y'all chose to create a specific variable for the dataPresent over possibly doing it inline I guess with the mapping of the data? 

This could totally be a nitpick so as Alex put in slack I'm gonna put `NIT` haha because this is totally valid just curious.

```suggestion
{data.length !== 0 ? data.map((list, index) => {<SingleList key={index} name={list.name} path={list.path} setListPath={setListPath} /> : null}
```

bbland1, line 12: It's awesome to actually see a list once I added the sharedList field to myself!

Was there a reason y'all chose to create a specific variable for the dataPresent over possibly doing it inline I guess with the mapping of the data? 

This could totally be a nitpick so as Alex put in slack I'm gonna put `NIT` haha because this is totally valid just curious.

```suggestion
{data.length !== 0 
? data.map((list, index) => {
<SingleList key={index} name={list.name} path={list.path} 
setListPath={setListPath} /> : null
}
```

zahrafalak: Thanks for your feedback! There isn't a very specific reason beyond just to improve readability and make the condition explicit. I agree that inlining could work too, especially for simpler cases—just wanted to make the logic clear and maintainable.

bbland1: That makes sense, and it's good to think about the maintainability of the code! 👍🏽

eternalmaha (approved):
Hey guys! This code looks very clean and readable! Nice job! 🫡. One I got curious about was the decision to use “data.length !== 0 “ . Negative lengths I believe are impossible haha, and this is really just semantics, but what do you guys think about the decision to check “data.length > 0 “ so it explicitly checks whether the length is greater than 0? I’m curious about your thoughts!

bbland1 (approved):

bbland1:
> Hey guys! This code looks very clean and readable! Nice job! 🫡. One I got curious about was the decision to use “data.length !== 0 “ . Negative lengths I believe are impossible haha, and this is really just semantics, but what do you guys think about the decision to check “data.length > 0 “ so it explicitly checks whether the length is greater than 0? I’m curious about your thoughts!

I do like the way that `data.length > 0` since they do the same check it is like a nicer read on my eyes but I do think both ways are valid and work. I can't think of any like performance benefits or issues on either.

zahrafalak:
Thanks for the kind words and great question! You're absolutely right—negative lengths are impossible, so checking data.length > 0 would work just as well and might even be a bit clearer. We went with data.length !== 0 to explicitly highlight that we're checking for non-empty data, but I'm totally open to using the > 0 approach for clarity. Appreciate the feedback!

zahrafalak:
> Hey guys! This code looks very clean and readable! Nice job! 🫡. One I got curious about was the decision to use “data.length !== 0 “ . Negative lengths I believe are impossible haha, and this is really just semantics, but what do you guys think about the decision to check “data.length > 0 “ so it explicitly checks whether the length is greater than 0? I’m curious about your thoughts!

Thanks for the kind words and great question! You're absolutely right—negative lengths are impossible, so checking data.length > 0 would work just as well and might even be a bit clearer. We went with data.length !== 0 to explicitly highlight that we're checking for non-empty data, but I'm totally open to using the > 0 approach for clarity. Appreciate the feedback!

src/views/Home.jsx
```diff
@@ -1,16 +1,24 @@
 import './Home.css';
+import { SingleList } from '../components';
 
 export function Home({ data, setListPath }) {
+	const dataPresent = data.length !== 0;
```
adidalal, line 5: One option is to inline this check and inverting it

So use: (!Array.isArray(data) || data.length == 0)

adidalal (approved):

src/views/Home.jsx
```diff
@@ -1,16 +1,24 @@
...
 			<p>
 				Hello from the home (<code>/</code>) page!
 			</p>
 			<ul>
-				{/**
-				 * TODO: write some JavaScript that renders the `lists` array
-				 * so we can see which lists the user has access to.
-				 */}
+				{dataPresent
+					? data.map((list, index) => (
```
tannaurus, line 13: Anytime you see a "else, null" in a ternary, you can do something like this instead

```
dataPresent && <MyComponent />
```

https://legacy.reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

tannaurus: I also love when variables are named with a touch more context, like something like `hasLists` instead of `dataPresent`. This is especially helpful in languages like JavaScript that don't have compile time type checking. Naming a variable in a way that communicates to the reader what this "data" is can go a long way in maintaining a healthy codebase 🚀

src/views/Home.jsx
tannaurus: if we're worried about the array being null, you can use the "nullish coalescing operator" to give the prop a default value

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
```
let list = null;

// assign a default at the top of our component
let listProp = list ?? [];
console.log(listProp) // logs "[]"
```

bbland1: I've seen the nullish coalescing operator before but didn't have a full grasp on how. Thanks for the docs and an example it makes so much more sense!

alex-andria (approved):
Great work overall on your first repo! 🥳🌟🌟 I loved reading the discussions, questions, and recommendations on this PR. I see there are a few potential areas to refactor if you decide to update those before merging to `main` - which is totally normal in PR reviews and what these are for 🙌. Nice job making the code simple and clean as well.

My suggestion for writing your next PR is to include the `Testing Steps / QA Criteria` like how Briana and Maha written it for their PR in Issue #1. This makes it easier for those of us testing the code in our local repos to know what changes to look/test for, especially as the issues get more complex throughout the cohort. Nice work Falak and Ross!

src/views/Home.jsx
zahrafalak: Love the recommendations and suggestions. Thanks so much for the feedback!

src/views/List.jsx
```diff
@@ -1,17 +1,16 @@
 import { ListItem } from '../components';
 
 export function List({ data }) {
+	const dataPresent = data.length !== 0;
```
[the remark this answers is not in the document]
zahrafalak: That is a great callout! We could definitely move this somewhere in a way that results in non repetitive code.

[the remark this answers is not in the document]
zahrafalak: That is a great callout! By my understanding, the 'data' prop passed onto these two components contain different information, even if the name of the prop is same. Do I understand that correctly? I guess that adds even more value to the suggestion above for using more contextual variable names like 'hasLists' or 'hasItems'.

RossaMania:
I'm so glad we were able to change the var names to provide context! The in-line && operator makes things even easier to read! And all the checks passed! Yay!

PR #22, opened 2024-08-19
adding a form layout to manage list

bbland1:

## Acceptance Criteria

UI-related tasks:
- [x] The `ManageList` view displays a form that allows them to enter the name of the item and select how soon they anticipate needing to buy it again. There should be 3 choices for this:
	- “Soon”, corresponding to 7 days
	- “Kind of soon”, corresponding to 14 days
	- “Not soon”, corresponding to 30 days
- [ ] The input that accepts the name of the item has a semantic `label` element associated with it
- [ ] The user can submit this form with both the mouse and the `Enter` key  
- [ ] When the user submits the form, they see a message indicating that the item either was or was not saved to the database.

Data-related tasks:
- [ ] The `console.log` in the `addItem` function in `src/api/firebase.js`  is replaced with a function that adds the new document to the Firestore database. That function will be imported from the `firebase/firestore` module.
- [ ]  The user’s soon/not soon/kind of soon choice is used to calculate `nextPurchasedDate`

## Type of Changes

### DIFF
--- a/src/views/ManageList.jsx
+++ b/src/views/ManageList.jsx
@@ -1,7 +1,30 @@
 export function ManageList() {
+	const handleSubmit = (e) => {
+		e.preventDefault();
+		console.log('hi');
+	};
+
 	return (
-		<p>
-			Hello from the <code>/manage-list</code> page!
-		</p>
+		<div>
+			<p>
+				Hello from the <code>/manage-list</code> page!
+			</p>
+			<form onSubmit={handleSubmit}>
+				<label htmlFor="item">Item Name</label>
+				<input id="item" type="text" />
+				<br />
+				{/* radio inputs for how soon to buy. */}
+				<input type="radio" id="soon" name="when-to-buy" value="7" />
+				<label htmlFor="soon">Soon</label>
+				<br />
+				<input type="radio" id="kind-of-soon" name="when-to-buy" value="14" />
+				<label htmlFor="kind-of-soon">Kind of soon</label>
+				<br />
+				<input type="radio" id="not-soon" name="when-to-buy" value="30" />
+				<label htmlFor="not-soon">Not soon</label>
+				<br />
+				<button type="submit">Submit Item</button>
+			</form>
+		</div>
 	);
 }


RossaMania:
Yay! It was great pair programming yesterday! The next step is passing that props object from managelist to firebase!!!

src/views/ManageList.jsx
```diff
@@ -1,7 +1,38 @@
...
+
 	return (
-		<p>
-			Hello from the <code>/manage-list</code> page!
-		</p>
+		<div>
+			<p>
+				Hello from the <code>/manage-list</code> page!
+			</p>
+			<form onSubmit={handleSubmit}>
```
bbland1, line 19: hey @RossaMania what do you think about using `react-hook-form` on our form? I know we got the base going and I think all in the form we have left to do it capture the state of the form to pass the correct parameters to the addItem. 

We can make sure validation is in the form either way (make sure they put in an item name and select an time option) but this may make it a bit quicker as we make future iterations, but I also could be jumping the gun. :sweat_smile:

https://react-hook-form.com/form-builder

https://www.npmjs.com/package/react-hook-form

bbland1: While I think it could still be an option as we add functionality I rescind my suggestion. Lol I don't think it is needed right now and will maybe cause some scope creep.

src/api/firebase.js
```diff
@@ -165,6 +165,19 @@ export async function addItem(listPath, { itemName, daysUntilNextPurchase }) {
 	const listCollectionRef = collection(db, listPath, 'items');
 	// TODO: Replace this call to console.log with the appropriate
 	// Firebase function, so this information is sent to your database!
+
+	//This saves item to database.
+	await setDoc(listCollectionRef, {
```
bbland1, line 170: I do think the `await setDoc` is correct. I know we were looking at when creating a list how it works. I think we can use the console.log in the return to build the info for the `setDoc` here at this call. 

```suggestion
await setDoc(listCollectionRef, {
		dateCreated: new Date(),
		// NOTE: This is null because the item has just been created.
		// We'll use updateItem to put a Date here when the item is purchased!
		dateLastPurchased: null,
		dateNextPurchased: getFutureDate(parseInt(daysUntilNextPurchase)),
		name: itemName,
		totalPurchases: 0,
	});
```

src/views/ManageList.jsx
```diff
@@ -1,7 +1,38 @@
+import { addItem } from '../api/firebase';
+import toast, { Toaster } from 'react-hot-toast';
+
 export function ManageList() {
+	const notify = () => toast.success('Item added to list!');
+
+	const handleSubmit = (e) => {
+		e.preventDefault();
```
bbland1, line 8:
```suggestion
e.preventDefault();
const itemInfoForm = new FormData(e.target)

console.log({itemName: itemInfoForm.get("item"), daysUntilNextPurchase: itemInfoForm.get("when-to-buy")}
```

RossaMania:
Yes. I like the idea of React Forms. That is what I was having trouble
finding last night! Some sort of React library for forms would be helpful.

RossaMania:
Confirmed in a chat: we don't need any special React Forms libraries or hooks. What we have is fine. If we do anything special, it'll come back to haunt all of us later. My words.

src/views/ManageList.jsx
```diff
@@ -1,7 +1,74 @@
...
+			<p>
+				Hello from the <code>/manage-list</code> page!
+			</p>
+			<form onSubmit={handleSubmit}>
+				<label>
+					Item: <input id="item" type="text" name="item" required />
+				</label>
+				<br />
+
+				<label htmlFor="when-to-buy">
```
bbland1, line 31: I added this option because I know we were trying to decide which we prefered and I wanted to see it. I do like the select drop down but they are both passing their values in the formData of the submit.

src/views/ManageList.jsx
```diff
@@ -1,7 +1,74 @@
...
 export function ManageList() {
+	const notify = () => toast.success('Item added to list!');
+
+	const handleSubmit = (e) => {
+		e.preventDefault();
+		const itemInfoForm = new FormData(e.target);
+
+		console.log({
+			itemName: itemInfoForm.get('item'),
+			daysUntilNextPurchaseSelect: itemInfoForm.get('when-to-buy-select'),
```
bbland1, line 13: When we pick which select we like we only need to keep one of them, but wanted to make sure the infor actually passed.

src/api/firebase.js
bbland1: I believe this will be `await addDoc` to get it to add!

src/views/ManageList.jsx
bbland1: I think I figured out my question about the listPath value (yes middle of the night after a migraine 😅) in `app.jsx` for the call that has the `/manage-list` we can add the list path value there to add it. I believe it is in a different route there already. (I'm on my phone so hard to check)

src/views/ManageList.jsx
```diff
@@ -1,7 +1,118 @@
...
+						name="item"
+						value={itemName}
+						onChange={handleItemNameTextChange}
+						required
+						aria-label="Enter the item name"
+					/>
+				</label>
+				<br />
+				<p>
+					When to buy: <br />
```
tannaurus, line 70: did we consider using a `fieldset` and `legend` here? The `fieldset` is a nice way of grouping fields that are alike, and the `legend` allows you to give that fieldset a "caption." It would result in a very similar result but would have the added benefit of following html semantics, which can go a long way for screen readers :)

src/views/ManageList.jsx
```diff
@@ -1,7 +1,118 @@
...
+					<label htmlFor="soon">
+						<input
+							type="radio"
+							id="soon"
+							name="when-to-buy"
+							value="7"
```
tannaurus, line 76: some food for thought that doesn't at all need to lead to any changes here, just something to consider as you're building

it's easy, and super common, to conflate "data" with "identifier", but that can get messy. We're doing that here by having `handleNextPurchaseChange` set this input's numerical value to state, then reading that state's current value to determine if this is checked or not (`checked={itemNextPurchaseTimeline === '7'}`).

In the work place, it's not uncommon to have product requirements start shifting under your feet. One day the product manager says these should be 7, 14, and 30 day increments, three weeks later you're told that they want to run an experiment where half the users see 7, 14, and 30 days, and the other half sees 1, 3, and 7 days, along with a "choose your own" input.

Designing with that kind of flexibility in mind can be a double edged sword, as it's best to not make something overly complex when you don't need to; no one after all is going to be changing these product requirements. 
That being said, avoiding conflating "data" with "identifier" can be an excellent way of finding a quick win in making an implementation more flexible. Instead of using the number of days as the "value", you could set the `field id` to state, then translate that `field id` to a given value when needed, like inside `handleSubmit` and the `aria-labels`. That translation could be done in a number of ways, but an easy trick is to declare a new object where the "key" is the field id and the "value" is the number associated with that field id. Then when you need to, you can just use the field id stored in state as the key to access the numerical value. Now, if you're asked to change `7` to `1`, or what ever your PM has in store for you that day, it can be a much easier change; you can even swap out that object with an entirely new set of values! `soon` can even be represented with the color `blue`, `kind-of-soon` with `orange`, what ever is asked of you 😄

src/views/ManageList.jsx
```diff
@@ -1,7 +1,118 @@
...
+		});
+
+		try {
+			await toast.promise(
+				addItem(listPath, {
+					itemName,
+					daysUntilNextPurchase: itemNextPurchaseTimeline,
+				}),
+				{
+					pending: 'Adding item to list.',
```
tannaurus, line 34: Something that is nearly universally enforced by a linter, but doesn't appear to be here, is consistency between single and double quotes. I notice some inconsistencies here, which isn't a big deal, but it's usually nice to see picking between one or the other (I'm team double quotes personally) and sticking to it

src/views/ManageList.jsx
```diff
@@ -1,7 +1,118 @@
...
+		e.preventDefault();
+
+		console.log('Submitting item:', {
+			itemName,
+			itemNextPurchaseTimeline,
+			listPath,
+		});
+
+		try {
+			await toast.promise(
```
tannaurus, line 28: One thing that wasn't a strictly outlined in the acceptance criteria, but something I personally think should be there, is post form submission for behavior. A successful form submission almost always results in the some indicator that it was successful (handled wonderfully by `toast` here) as well as clearing the values that were previously there. On the flip side, if we were unable to submit the form, it's best practice to leave the values there so that the user can re-submit their request

tannaurus:
Left behind some thoughts, but this LGTM! 🚀  Great work!

src/views/ManageList.jsx
```diff
@@ -1,7 +1,118 @@
...
+
+	const handleItemNameTextChange = (e) => {
+		setItemName(e.target.value);
+	};
+
+	const handleNextPurchaseChange = (e) => {
+		setItemNextPurchaseTimeline(e.target.value);
+	};
+
+	const handleSubmit = async (e) => {
```
zahrafalak, line 18: Suggestion: It would be beneficial to reset the state values after a successful form submission, so the form fields are cleared and ready for new input.

src/views/ManageList.jsx
```diff
@@ -1,7 +1,118 @@
-export function ManageList() {
+import { useState } from 'react';
+import { addItem } from '../api/firebase';
+import toast, { Toaster } from 'react-hot-toast';
+
+export function ManageList({ listPath }) {
+	console.log('listPath:', listPath);
+	const [itemName, setItemName] = useState('');
+	const [itemNextPurchaseTimeline, setItemNextPurchaseTimeline] = useState('');
```
zahrafalak, line 8: It would be beneficial to reset the state values after a successful form submission, so the form fields are cleared and ready for new input.

src/views/ManageList.jsx
```diff
@@ -1,7 +1,118 @@
...
+	};
+
+	const handleNextPurchaseChange = (e) => {
+		setItemNextPurchaseTimeline(e.target.value);
+	};
+
+	const handleSubmit = async (e) => {
+		e.preventDefault();
+
+		console.log('Submitting item:', {
```
zahrafalak, line 21: Suggestion: You might want to remove or comment out the console.log statements before finalizing your code. They’re super helpful during development and debugging, but in production, they can clutter the console and might even reveal sensitive information.

src/views/ManageList.jsx
```diff
@@ -1,7 +1,118 @@
...
+							value="30"
+							required
+							onChange={handleNextPurchaseChange}
+							checked={itemNextPurchaseTimeline === '30'}
+							aria-label="Set buy to not soon, within 30 days"
+						/>
+						Not soon
+					</label>
+				</p>
+				<button type="submit">Submit Item</button>
```
zahrafalak, line 113: Praise: Love the use of aria labels to increase the accessibility of the input fields! 
Suggestion: We could also have an aria label for the interactive button element. This provides clarity, especially in forms where the purpose of a button might not be immediately clear to all users.

src/views/ManageList.jsx
bbland1: Yes! Thank you for making the reminder too. We had some "it worked on my computer" moments when we couldn't meet up so they have played their parts 😂

src/views/ManageList.jsx
bbland1: Yeah I could of sworn it was getting linted but here they are inconsistent 😂

src/views/ManageList.jsx
bbland1: Thank you for this! It was exactly what we were going for but couldn't find the right tags!

RossaMania:
Thanks everybody for the sweet feedback! This is so cool! We were able to remove redundant console logs, use key-value pairs in an object for a "purchaseTimeline". We were also able to apply some aria-labels! We tried to change single quotes to double quotes, but it kept reverting to single quotes on some! :D

src/views/ManageList.jsx
bbland1: @tannaurus This is something awesome to think about and keep in mind when coding! I think we were able to come up with a change on getting this to work! 

This may be a overly complex for this current moment, but with making the object and changing the values on the inputs I notice it's a lot of spots to hit if you need to change the key names. 

So more just as a curiosity would this be more flexible if we make our new `purchaseTimelines` object like an array of objects that have like the id, label and value information for the inputs to map over those and dynamically make the radio buttons?

src/views/ManageList.jsx
```diff
@@ -1,7 +1,126 @@
...
+					},
+					error: () => {
+						return `${itemName} failed to add to your list. Please try again!`;
+					},
+				},
+				{
+					style: {
+						minWidth: '250px',
+					},
+				},
```
zahrafalak, line 49: This might potentially be a nit pick suggestion, but is there a way for the style element to not be present in the jsx file, just so styling elements are all saved separately?

bbland1: If I had to take a full guess probably 😅, this was following documentation from the `react-hot-toast` how to set it up with the options. So it would just be a double check on if this is specific to the use of the toast or just an option in the usage. 🤔

src/views/ManageList.jsx
```diff
@@ -1,7 +1,126 @@
...
+							type="radio"
+							id="soon"
+							name="when-to-buy"
+							value="soon"
+							required
+							onChange={handleNextPurchaseChange}
+							checked={itemNextPurchaseTimeline === 'soon'}
+							aria-label="Set buy to soon, within 7 days"
+						/>
+						Soon -- Within 7 days!
```
tannaurus, line 88: In order for us to obtain the flexibility of `purchaseTimelines`, we'd have to reference that here (and in the aria-label) with something like this for the label text:

> "{ `Soon -- Within ${purchaseTimelines["soon"]}!` }"

And something like this for the aria-label

> `Set buy to soon, within ${purchaseTimelines["soon"]} days`

tannaurus: Otherwise, we can't change the values in `purchaseTimelines` without making changes throughout this file 😄

src/api/firebase.js
```diff
@@ -163,14 +164,13 @@ export async function shareList(listPath, currentUserId, recipientEmail) {
...
-	// Firebase function, so this information is sent to your database!
-	return console.log(listCollectionRef, {
+
+	await addDoc(listCollectionRef, {
 		dateCreated: new Date(),
 		// NOTE: This is null because the item has just been created.
 		// We'll use updateItem to put a Date here when the item is purchased!
 		dateLastPurchased: null,
-		dateNextPurchased: getFutureDate(daysUntilNextPurchase),
+		dateNextPurchased: getFutureDate(parseInt(daysUntilNextPurchase)),
```
tannaurus, line 173: Now that `purchaseTimelines` are integers, we shouldn't have to `parseInt` anymore 😄

src/views/ManageList.jsx
```diff
@@ -1,7 +1,126 @@
...
+	};
+
+	const handleNextPurchaseChange = (e) => {
+		setItemNextPurchaseTimeline(e.target.value);
+	};
+
+	const handleSubmit = async (e) => {
+		e.preventDefault();
+
+		const daysUntilNextPurchase = purchaseTimelines[itemNextPurchaseTimeline];
```
tannaurus, line 26: since JavaScript is so loosy goosy, we probably should be checking if `purchaseTimelines[itemNextPurchaseTimeline]` exists before continuing, or at least verify that our database won't be put into a bad place if `purchaseTimelines[itemNextPurchaseTimeline]` returns `undefined`

```
const times = {
  "now": 0,
  "sooner": 3,
  "soon": 7,
}

const time = times["now"];

const typo_time = times["yesterday"];

console.log(time, typo_time);
```

We want to do this because we if accidentally let this slip by now, we'd have a problem where our database contains data the client probably doesn't exist; which would lead to errors down the line. If `daysUntilNextPurchase` is undefined, and the database lets it be inserted (assuming that will succeed, worth checking!) when it comes time to render our list, we may rightfully assume a day is stored in that field, try to parse it, and get an error. Now image the person that creates this bad list shares it with other users, now they're impacted too 💣 Normally you'd expect the server to do this kind of validation, but I'm not sure how firebase will behave here.

tannaurus: Using a set of constants here can also cut down on the chance of having typos like this. You can declare your times at the top of the file and then reference them in `purchaseTimelines`

```
const SOON = "soon";
const KIND_OF_SOON = "kindOfSoon";
const NOT_SOON = "notSoon"

const purchaseTimelines = {
  [SOON]: 7,
  ... etc
};

# In your render function 

... snip
checked={itemNextPurchaseTimeline === SOON}
... snip
```

adidalal: I know we don't use TypeScript here, but it has native support for enums, which are also helpful when it comes to definitions that get reused - possibly something to 👀 into at some point

tannaurus (approved):

src/views/ManageList.jsx
tannaurus: 100%!

src/views/ManageList.jsx
```diff
@@ -1,7 +1,126 @@
...
+
+		try {
+			await toast.promise(
+				addItem(listPath, {
+					itemName,
```
tannaurus, line 31: I left a similar comment on PR #23 , we might want to try to try and validate this input 😄 

https://github.com/the-collab-lab/tcl-77-smart-shopping-list/pull/23/files#r1730100136

src/views/ManageList.jsx
tannaurus: I think that's a great idea! It would also encourage us to break out each radio button into its own component. One prop could tell that component which "field_id" it represents, along with what "field_id" is current set in the parent's state

I'm personally ok with leaving it as-is or taking it to that level; my view is that my role here is to give advice on the implementation as it iterates. I don't mind doing that over and over again as these ideas develop, but I'd defer to your ability to make those changes to be the deciding factor here 😄 I think the collab lab is a great place to experiment with these kind of design philosophies and I'm happy to give feedback on those experiments if/when you've got the time 🚀

src/views/ManageList.jsx
adidalal: also as a pointer - if you're doing the same thing multiple times...might be a sign to make it a utility function :)

eternalmaha (approved):
The code is so well structured and readable! I saw some really great suggestions that take future implementations into account! Possible interesting take could also have been to dynamically render the input elements through mapping a revised purchaseTimeLines object! In this case, there's only three options, so it's not necessary at all but if the project expanded, it could be a useful code revision to increase scalability and reduce lines of code!

bbland1 (commented):
> The code is so well structured and readable! I saw some really great suggestions that take future implementations into account! Possible interesting take could also have been to dynamically render the input elements through mapping a revised purchaseTimeLines object! In this case, there's only three options, so it's not necessary at all but if the project expanded, it could be a useful code revision to increase scalability and reduce lines of code!  

I think this is something I would definetly like to look at trying in a refactor or revision, like you mention I think it would bring some good things to the code! One of Tanner's comments made me think about doing an array as well! I know it isn't neccary for this PR and maybe not for the project but I'd like to give it a try maybe on a later week because at this moment I'm not sure how to do the refactor for it but I think it would be really nice to get the radio button into a component and use the array to may over and dynamically make them.

PR #27, opened 2024-08-26
Restrict Navigation Bar Links Based on User Authentication Status

### LINKED ISSUE #26: Restrict Navigation Bar Links Based on User Authentication Status
#### Description
We utilize Google Auth in the app to validate users are seeing their own lists and items, and it's awesome. When working on issue #5 we noticed that in some troubleshooting that while the user is needed to access the lists and the form on the manage list page a user could navigate to their like base pages. It caused some "it works on my computer" moments with trying to just use the form without signing in first since you can navigate to the page. It seems that making sure the nav bar utilizes the user object somehow like the signin/sign out button would be a better flow for users.

#### Acceptance Criteria
- [x] The sign in button is on the homepage when no user is logged in
- [x] The sign out button is in the nav bar when a user is logged in and there is no sign in button
- [x] The nav bar is only visible when a user is logged in
- [x] Prevent manual navigation to routes that a user has to be logged into see

#### Screenshots
The screenshots are on mobile because I was just making sure it worked there too when it was noticed again after the issue. I can click those links and see those pages. The parts that are protected by the user aren't there but I think it may be nice to have those parts of the navbar not show up if no one is logged in since the are specific to the app.

https://github.com/user-attachments/assets/6379de02-3372-4e7b-a357-2f6153089377

#### Files
To make the changes the files affected would be 🤔:
- `App.tsx`
- `Layout.tsx`
- `useAuth.tsx`?

#### To Reproduce
When navigating to the application if you are not logged in you will be able to click all the navlinks because they will display and it will say the "hi from this page" on each of them.

#### Tasks
- [x] In the `Layout` component in the `App.tsx` the user prop could be added and pass the user info the app gets from calling the `useAuth` function, and use that for th user checks instead of calling useAuth again.
- [x] Move the sign in button out of the nav bar so on the home page it can display on the page when a user is not signed in. Maybe into the main tag right now and change the check to be something like `!!user ? null : <SignInButton />` so that if the user isn't logged the button will be the sign in version.
- [x] Have the NavBar show based on if a user is sign in out not by using the same user check that is currently on the sign in button but for the nav bar. Surrounding the navbar in a user check with `user ? NAV BAR : null`. The signout button in the nav bar would stay but the check would change to something like `!!user ? <SignOutButton /> : null` 
- [x] The sign out button in `useAuth.tsx` would probably need to be updated to redirect/ navigate the page back to the home page if a user is on the list or manage list page when they hit the signout button.

bbland1:
## Description

This code change improves the user experience to reduce confusion on function usability when signed in and out of the application. It dynamically adapts the navbar to only show when a user is logged in and prevents "crafty" users from manually navigating to protected routes and unestablished routes by adding redirects back to the homepage. 

There was some own confusion of being able to navigate to pages when working on other issues and not understanding why I wasn't seeing what I thought there should be only to realize I wasn't logged in. If I was having that issue a new user could land on the application and attempt to navigate and not realize they are missing out because they aren't logged in. We want to provide a good user experience that directs people to the proper use of the application. 

Stumbling upon this issue did make me realize that authentication while handled by firebase still has a UI component that we can think about other than the direct login button. Having the UI change as we have different user states allows for a subconscious understanding (that is fostered by the rest of the web) of how our application should work from the moment of signup.

## Related Issue

Closes #26

## Acceptance Criteria

- [x] The sign in button is on the homepage when no user is logged in
- [x] The sign out button is in the nav bar when a user is logged in and there is no sign in button
- [x] The nav bar is only visible when a user is logged in
- [x] Prevent manual navigation to routes that a user has to be logged into see

## Type of Changes

|  ✓    | 💯 Enhancement     |
|  ✓    | ♻️ Refactor |
|  ✓    | 🐛 Bug Fix |

### Before

https://github.com/user-attachments/assets/4901eea2-bbf4-4ac0-bd2c-abfbadb5f2e3

### After

https://github.com/user-attachments/assets/145e2aec-ea31-4bc6-a153-8fd579c7cdcd

- From your terminal, pull down this branch with `git pull origin bb/navbar-auth-status` and check that branch out with git checkout `bb-/navbar-auth-status`
- Then `npm ci` to install the newly added dependencies locally and `npm start` to launch the app.
- Open a second terminal to run the typescript compiler using ` tsc --noEmit --watch` to see any type issue.
- If still signed in from previous usage of the app, make sure to sign out and sign back in. 
- When signed out you should only see the sign in button on the homepage with the "hello from the homepage message".
- In the URL bar add `/list` to the end and see that the page will automatically redirect to the homepage if no user is signed in. The same will happen if `/manage-list` is added.
- After seeing the redirect of non-logged in users to the homepage if manually changing the page click the `sign in` button.
- When logged in the nav bar will appear back at the bottom of the page and can be used as it previously was.
- Hit `sign out` and get redirected back to the home page on successful log out.


src/api/useAuth.tsx
bbland1, line 1: In the Office hours we talked about maybe having the useAuth hook handle the redirect so the call of useAuth in the App.tsx we don't have to ever pass null. Would that mean like doing a user check here that resturns the user object and if not this would use the useNavigate to navigate to home page? 

Also I imagine that changes the protected routes handling but I'm not sure I am in the right line of thought.

src/views/Layout.tsx
```diff
@@ -25,25 +18,28 @@ export function Layout() {
 				</header>
 				<main className="Layout-main">
 					<Outlet />
+					{!!user ? null : <SignInButton />}
```
tannaurus, line 21: Rule of thumb: If you ever see `null` in a ternary, you don't need a ternary

```suggestion
					{user ?? <SignInButton />}
```

src/App.tsx
```diff
@@ -2,12 +2,21 @@ import React from "react";
...
+import { Home, Layout, List, ManageList, PageNotFound } from "./views";
 
 import { useAuth, useShoppingListData, useShoppingLists } from "./api";
 
 import { useStateWithStorage } from "./utils";
 
+import { ProtectRoute } from "./components";
+
+/**
+ * Putting Toaster at the top level of the App allows
```
tannaurus, line 14: I really like this change!

src/views/Layout.tsx
```diff
@@ -25,25 +18,28 @@ export function Layout() {
...
-						<NavLink
-							to="/manage-list"
-							className="Nav-link"
-							aria-label="Manage List"
-						>
-							Manage List
-						</NavLink>
-					</div>
-				</nav>
+				{user ? (
```
tannaurus, line 23: Same here 😄 
```suggestion
				{user && (
```

src/App.tsx
```diff
@@ -43,19 +52,29 @@ export function App() {
...
 						}
 					/>
-					<Route path="/list" element={<List data={data} />} />
-					<Route
-						path="/manage-list"
-						element={<ManageList listPath={listPath} />}
-					/>
+
+					{/* protected routes */}
+					<Route element={<ProtectRoute user={user} redirectPath="/" />}>
```
tannaurus, line 66: I'm having a bit of a hard time with this one. On one hand, I see that just about everyone on the www suggests using `Outlet` (or something akin to it). On the other hand, it still forces you to content with `user` being `User | null` anywhere you invoke the `useAuth` hook. That suggests to me we might be using Outlet wrong here, and frankly I'd go to bat against the www on this one too 😅 

I think `Outlet` is fantastic for inserting components across a number of views, like our nav bar, which is how the [React Router documentation suggests its used](https://reactrouter.com/en/main/components/outlet). Because we're forced to content with the types, I think we should focus our solution there instead by recycling a common function naming convention: prefixing with `get` and `find` (`useGetUser` / `useFindUser`). Commonly, `get` isn't expected to fail, it throws an error if it doesn't find what its looking for, but `find` can come back empty handed without it being considered an error. Mapping that idea to our use case:
- Rename `useAuth` to `useFindUser` -> it isn't considered an error if a user isn't found, so it returns `User | null`
- `useGetUser` would use `useFindUser` under the hood to get the user and redirect to `/` if `user` is `null`

Now, in places like the nav bar where we want to render two separate components based on if `user` is defined, we can use `useFindUser`, and in places where we need the component to be restricted to just authenticated users, we can use `useGetUser`. 

Once we've got those types situated, I think we should still still keep `ProtectedRoute`, but maybe shift it up to be wrapping all of the routes that are "protected", instead of wrapping them individually. That will sort out the types + the chance there's a view that doesn't need `user` but isn't meant to be seen by non-users

src/views/Layout.tsx
bbland1: This is a good rule of thumb to know and to remember that the `??` exists. I definitely looked at is was like 🤔 but I couldn't recall the other option or find the words for the google search!

src/App.tsx
bbland1: 😅 I am not forcing any battles with the www 😂. After office hours on Monday and our talk about how to handle not passing props down a never ending chain, I had a feeling this area would get some tweaks lol. 

This is really helpful to know about the why of the naming convention, I had see both use and kind of was seeing a difference but didn't fully know this! 

So now we have `useAuth` call at the top in `App` and it gets passed along basically, but with this would `Layout` and `App` now have their own calls of the needed function in them? Like `App` using `useGetUser` and passing that to the `ProtectedRoute` and other routes, while `Layout` would no longer get a user prop because it can call `useFindUser` because we need to handle the view of the navbar based on the null vs user?

src/views/Layout.tsx
bbland1: when trying the suggestion it causes a type error because it seems like user isn't a react node.

would something other than this work?

```
{user ? <></> : <SignInButton />}

```

with this it works but it seems odd to just have a fragment to get it to work.

RossaMania:
I like it! It makes sense, and can prevent future headaches later! We're already checking for a user to be logged in in our issue for this week.

src/App.tsx
RossaMania: Commenting because I agree with using a ProtectedRoute. We can have it just wrap the routes using Auth.

src/api/useAuth.tsx
```diff
@@ -54,7 +75,23 @@ export const useAuth = () => {
...
+
+export const useGetUser = () => {
+	const { user } = useFindUser();
+	const navigate = useNavigate();
+
+	useEffect(() => {
+		if (user === null) {
+			navigate("/", { replace: true });
+		}
+	}, [user, navigate]);
```
bbland1, line 94: So right now this doens't actually only return the user it is just a differnet version of the `useFindUser` method. I have tried a few things but when I throw an error the whole app will crash just always hitting that error. 

I was trying to just put it in the `protectedRoute` and get the full change to work but I am a bit stuck. 😅 I have seemed to notic right now when I did a test console.log of it in `protectedRoute` that there seems to be 3 renders happening. I put it in a `useEffect` and did the console.log and still got those 3 console.logs with 2 being null and 1 being the data we need. So I do notice that with that it will automatically redirect because the first time it is `null`. It may be a needing to manage state in `useGetUser` like in `useFindUser` but I haven't quite determined that or what direction should it go.

Any suggestions?

bbland1:
With commit [36e2def](https://github.com/the-collab-lab/tcl-77-smart-shopping-list/pull/27/commits/36e2defb8c38f517a8456495c94e33c6ce773d24) I wanted to confirm that the `getUser` function added to the `ProtectedRoute` component used for context to pass the `user` to the routes we are protecting can't be used in a route not protected. I added it to the `Home` component and the app did crash. With an issue of being unable to destructure the user information, and wouldn't load the app.

https://github.com/user-attachments/assets/f00c3901-81db-43b2-9f34-bfebb16da759

package-lock.json
```diff
@@ -7951,9 +7951,9 @@
 			}
 		},
 		"node_modules/micromatch": {
-			"version": "4.0.7",
-			"resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.7.tgz",
-			"integrity": "sha512-LPP/3KorzCwBxfeUuZmaR6bG2kdeHSbe0P2tY3FLRU4vYrjYz5hI4QZwV0njUx3jeuKe67YukQ1LSPZBKDqO/Q==",
+			"version": "4.0.8",
```
tannaurus, line 7954: we should be able to revert this 👀

src/api/useAuth.tsx
```diff
@@ -20,22 +22,41 @@ export const SignInButton = () => (
...
+	return (
+		<button
+			type="button"
+			onClick={() => {
+				auth
+					.signOut()
+					.then(() => {
+						navigate("/", { replace: true });
+					})
+					.catch((error) => {
```
tannaurus, line 37: Great idea with handling this!

src/api/useAuth.tsx
```diff
@@ -20,22 +22,41 @@ export const SignInButton = () => (
...
+export const SignOutButton = () => {
+	const navigate = useNavigate();
+
+	return (
+		<button
+			type="button"
+			onClick={() => {
+				auth
+					.signOut()
+					.then(() => {
```
tannaurus, line 34: You shouldn't have to do this because `ProtectedRoute` is doing its job very well 😄 it can be a double edged sword doing this imo. On one hand, if we weren't able to `subscribe` to events like we can with Firebase, you'd _have_ to do this; in fact, I generally prefer seeing very "direct" actions like this because it makes it very easy to reason with the application. That being said, because the app doesn't _need_ this code to achieve the functionality, it can actually work against our goal of making this action "direct"; having this here suggests that this could be changed to change the sign out direct functionality, but that's not the case: it's adding complexity. You should be safe to remove this 🙏

src/api/useAuth.tsx
```diff
@@ -54,6 +75,9 @@ export const useAuth = () => {
 			setUser(user);
 			addUserToDatabase(user);
 		});
+
+		// Cleanup the subscription when the component unmounts
+		return () => unsubscribe();
```
tannaurus, line 80: Great catch!

src/views/PageNotFound.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export function PageNotFound() {
+	const navigate = useNavigate();
+
+	useEffect(() => {
+		const timer = setTimeout(() => {
+			navigate("/", { replace: true });
+		}, 3000);
+
+		return () => clearTimeout(timer);
```
tannaurus, line 12: Nice touch with the use of `navigate` here, and great job thinking about clearing the timeout!

src/components/NavBar.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { NavLink } from "react-router-dom";
+import { SignOutButton } from "../api";
+
+import "./NavBar.css";
+
+export function NavBar() {
```
tannaurus, line 7: _love_ that this is a new component. Maybe it could called `AuthenticatedNavBar` just to indicate where it should be used? That would open the door for us to eventually have an `UnauthenticatedNavBar`, which might be nice at the end of this project where we're may be interested in adding a "home page" that describes what this app is doing

src/App.tsx
```diff
@@ -1,13 +1,22 @@
 import React from "react";
 
-import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
+import { Routes, Route } from "react-router-dom";
 
-import { Home, Layout, List, ManageList } from "./views";
```
tannaurus, line 5: Last piece of "food for thought" here comes from experience in working in very large React apps: eventually the directory structure of directories like "views" can be very confusing. I find a great solution to this is to have a directory like "views" have two sub directories: something like "authenticated" and "unauthenticated". Our `List`, `ManageList`, and others could live in the new `authenticated` directory, and the new `PageNotFound` and potentially others could live in the `unauthenticated` directory

When making changes like that, there's two approaches: one is preferred, but sometimes the other is unavoidable. The preferred option here, that would work well for this work, would be to create a new PR that is branched off of this one (or wait until this merges and branch it off of main), and do the file movements in that PR. Something important with this path would be to *only* move the files, don't change anything else. This allows you to say in your MR "just moving things around" , which makes it infinitely easier for reviewers, as well as your teammates who will likely have to deal with merge conflicts.

The other solution would be to do the changes here, which I would advise you _always_ try to avoid because it can make the diffs seem much larger than they really are and make reviewing the "real" changes harder.

tannaurus (approved):

tannaurus:
Exceptional work @bbland1 ! I love these changes.

src/components/ProtectedRoute.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { User } from "../api";
+
+interface Props {
+	user: User | null;
+	redirectPath: string;
+}
+
+type ProtectedRouteProps = { user: User };
+
+export function ProtectRoute({ user, redirectPath }: Props) {
```
tannaurus, line 12: Now that this is in place, we probably want to scan the codebase to remove all `User | null` where its no longer needed, and use the new `getUser` function where we previously used the old one

package-lock.json
bbland1: Yeah not sure how that got through 😅. I know I ran the npm audit when I was installing pacakges after the the TS update but I don't remember actually running npm audit fix so I don't really know what happened there.

bbland1: I'm happy to revert this, I'm not fully sure how. I think from what I'm googling I need to find the hash where it is the original and check out and commit again. Is that the correct track?

I was also looking at what is on main right now and I see it is also 4.0.8, will it cause any issues if I revert and then merge?

src/components/ProtectedRoute.tsx
bbland1: I think with fixing my merge conflicts of the main after the other two PRs got merged I call all the `User | null` that were suppose to be there any more. I think the only place I see it now is in places where the `protectedRoute` isn't used like the `App.tsx`, `Home.tsx`, `Layout.tsx` and in `protectedRoute`.

src/App.tsx
bbland1: This is really good info. I was starting to feel like maybe "reorganization" was on the forefront, but with more files in play I didn't think it was something that would typically be attached to a PR where code has been changed. So knowing that there is multiple approaches, and a preferred method, is really nice. 

I'm gonna take the create PR off of this one and move because I just also prefer that. It seems cleaner and easier to deal with.

PR #28, opened 2024-08-28
#6, feat: Add form that allows user to invite another user to list!

### LINKED ISSUE #6: 6. As a user, I want to be able to invite others to an existing shopping list
## Summary

We want to allow users to invite others to existing shopping lists so they can manage them with friends or family. This would take the following steps:

1. User enters an existing user's email into a form
2. The `shareList` function in `src/api/firebase.js` is called with the appropriate inputs

Please note that an existing user is a user that has signed into the app before.

## Acceptance criteria

- [ ]  The `ManageList` view shows a form that allows the user to enter an email to invite an _existing_ user to a list, in addition to the form that allows them to add items to that list.
- [ ]  The input that accepts the email  has a semantic `label` element associated with it
- [ ]  The user can submit this form with both the mouse and the `Enter` key
- [ ]  If the other user exists, the user is alerted that the list was shared
- [ ]  If the other user _does not_ exist, the user is shown an error message that explains the problem

## Notes

- You will know your `<label>` is correctly implemented if you click on the label and keyboard focus moves to the related input. Refer to [the MDN docs on `<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) for more info.
- Users will be able to submit the form with the `Enter` key if you listen for the right JavaScript event on the right element! If you're adding an `onKeyDown` listener anywhere, you might be overcomplicating things.
- Check the `shareList` function in `src/api/firebase.js` to get an idea of how you will know if a list was shared.

RossaMania:

## Description

We want to allow users to invite others to existing shopping lists so they can manage them with friends or family. We created a form in the ManageList view that allows the user to enter an email to invite an existing user to a list. It's in addition to the form that allows them to add items to that list. The user can submit this form with both the mouse and the Enter key. If the other user exists, the user is alerted that the list was shared with a React Hot Toast notification. If the other user does not exist, the list will not be shared and user will be notified!

This uses event listeners, and the shareList function from the firebase.ts!

We learned that it is important to keep it simple. While it may feel like we're reinventing the wheel even though we went from JavaScript to TypeScript in week 3, this is still a CRUD app. It was so cool to go through the different files and see how all the different arguments and props were passed! 

## Related Issue
closes #6 

## Acceptance Criteria
 The ManageList view shows a form that allows the user to enter an email to invite an existing user to a list, in addition to the 
 form that allows them to add items to that list.
  The input that accepts the email has a semantic label element associated with it
  The user can submit this form with both the mouse and the Enter key
  If the other user exists, the user is alerted that the list was shared
  If the other user does not exist, the user is shown an error message that explains the problem

## Type of Changes

`enhancement`

### Before

image

### After

image

1. Sign in to app.
2. Pick a list from the Home page.
3. Click on "Manage List" 
4. Enter email address in the text field to the left of the "Invite User" button.
5. Click on "Invite User" button or press Enter.
6. Wait for the notification! The notification will either be green and show the email address and the message that the user was successfully invited to your list, or it will be red, show the email address, and the message that the user was not added to list!

### DIFF
--- a/src/views/ManageList.tsx
+++ b/src/views/ManageList.tsx
@@ -102,7 +102,8 @@ export function ManageList({ listPath }: Props) {
 		}
 
 		if (currentUser === null) {
-			toast.error("You are not logged in! Cannot invite.");
+			toast.error(`You are not logged in! Cannot invite.`);
+
 			return;
 		}
 
@@ -198,13 +199,14 @@ export function ManageList({ listPath }: Props) {
 							Submit Item
 						</button>
 					</form>
-					<Toaster />
+
 					<form onSubmit={(e) => handleInvite(e, listPath)}>
-						<label htmlFor="receipient-email">
+						<label htmlFor="recipient-email">
+							Recipient Email:
 							<input
-								id="receipient-email"
+								id="recipient-email"
 								type="email"
-								name="receipient-email"
+								name="recipient-email"
 								value={emailName}
 								onChange={handleEmailNameChange}
 								required
--- a/src/views/ManageList.tsx
+++ b/src/views/ManageList.tsx
@@ -1,7 +1,9 @@
 import { ChangeEvent, FormEvent, useState } from "react";
-import { addItem } from "../api/firebase";
+import { addItem, shareList } from "../api/firebase";
 import { validateTrimmedString } from "../utils";
 import toast, { Toaster } from "react-hot-toast";
+
+import { useAuth } from "../api/useAuth";
 
 interface Props {
 	listPath: string | null;
@@ -20,10 +22,13 @@ const purchaseTimelines = {
 };
 
 export function ManageList({ listPath }: Props) {
+	const { user: currentUser } = useAuth();
+
 	const [itemName, setItemName] = useState("");
 	const [itemNextPurchaseTimeline, setItemNextPurchaseTimeline] = useState(
 		PurchaseTime.soon,
 	);
+	const [emailName, setEmailName] = useState("");
 
 	const handleItemNameTextChange = (e: ChangeEvent<HTMLInputElement>) => {
 		setItemName(e.target.value);
@@ -31,6 +36,10 @@ export function ManageList({ listPath }: Props) {
 
 	const handleNextPurchaseChange = (changed: PurchaseTime) => {
 		setItemNextPurchaseTimeline(changed);
+	};
+
+	const handleEmailNameChange = (e: ChangeEvent<HTMLInputElement>) => {
+		setEmailName(e.target.value);
 	};
 
 	const handleSubmit = async (
@@ -73,6 +82,43 @@ export function ManageList({ listPath }: Props) {
 			);
 		} catch (error) {
 			console.error("Failed to add item:", error);
+		}
+	};
+
+	const handleInvite = async (
+		e: FormEvent<HTMLFormElement>,
+		listPath: string,
+	) => {
+		console.log("Button clicked! Inviting user!");
+		e.preventDefault();
+
+		const trimmedEmailName = validateTrimmedString(emailName);
+
+		if (!trimmedEmailName) {
+			toast.error(
+				"Oops! Email cannot be empty or just spaces. Please enter a valid email.",
+			);
+			return;
+		}
+
+		if (currentUser === null) {
+			toast.error("You are not logged in! Cannot invite.");
+			return;
+		}
+
+		try {
+			await toast.promise(shareList(listPath, currentUser, trimmedEmailName), {
+				loading: "sharing list with existing user",
+				success: () => {
+					setEmailName("");
+					return `${emailName} successfully invited to your list!`;
+				},
+				error: () => {
+					return `${emailName} Oops! Failed to invite to your list. Please try again!`;
+				},
+			});
+		} catch (error) {
+			console.error("Oops! Failed to invite user:", error);
 		}
 	};
 
@@ -150,6 +196,24 @@ export function ManageList({ listPath }: Props) {
 						</fieldset>
 						<button type="submit" aria-label="Add item to shopping list">
 							Submit Item
+						</button>
+					</form>
+					<Toaster />
+					<form onSubmit={(e) => handleInvite(e, listPath)}>
+						<label htmlFor="receipient-email">
+							<input
+								id="receipient-email"
+								type="email"
+								name="receipient-email"
+								value={emailName}
+								onChange={handleEmailNameChange}
+								required
+								aria-label="Enter the user name to share list"
+								aria-required
+							/>
+						</label>
+						<button type="submit" aria-label="Add item to shopping list">
+							Invite User
 						</button>
 					</form>
 					<Toaster />
--- a/src/views/ManageList.tsx
+++ b/src/views/ManageList.tsx
@@ -1,7 +1,7 @@
 import { ChangeEvent, FormEvent, useState } from "react";
 import { addItem, shareList } from "../api/firebase";
 import { validateTrimmedString } from "../utils";
-import toast, { Toaster } from "react-hot-toast";
+import { toast } from "react-hot-toast";
 
 import { useAuth } from "../api/useAuth";
 
@@ -218,7 +218,6 @@ export function ManageList({ listPath }: Props) {
 							Invite User
 						</button>
 					</form>
-					<Toaster />
 				</>
 			)}
 		</div>
--- a/src/views/ManageList.tsx
+++ b/src/views/ManageList.tsx
@@ -1,7 +1,7 @@
 import { ChangeEvent, FormEvent, useState } from "react";
 import { addItem, shareList } from "../api/firebase";
 import { validateTrimmedString } from "../utils";
-import { toast } from "react-hot-toast";
+import toast from "react-hot-toast";
 
 import { useAuth } from "../api/useAuth";
 
@@ -112,10 +112,10 @@ export function ManageList({ listPath }: Props) {
 				loading: "sharing list with existing user",
 				success: () => {
 					setEmailName("");
-					return `${emailName} successfully invited to your list!`;
+					return `Successfully invited ${emailName} to your list!`;
 				},
 				error: () => {
-					return `${emailName} Oops! Failed to invite to your list. Please try again!`;
+					return `Oops! Failed to invite ${emailName} to your list. Please verify correct email!`;
 				},
 			});
 		} catch (error) {
@@ -208,12 +208,14 @@ export function ManageList({ listPath }: Props) {
 								type="email"
 								name="recipient-email"
 								value={emailName}
+								placeholder="Enter e-mail address. . ."
 								onChange={handleEmailNameChange}
 								required
 								aria-label="Enter the user name to share list"
 								aria-required
 							/>
 						</label>
+						<br />
 						<button type="submit" aria-label="Add item to shopping list">
 							Invite User
 						</button>
--- a/src/views/ManageList.tsx
+++ b/src/views/ManageList.tsx
@@ -1,9 +1,9 @@
 import { ChangeEvent, FormEvent, useState } from "react";
-import { addItem, shareList } from "../api/firebase";
+import { addItem } from "../api/firebase";
 import { validateTrimmedString } from "../utils";
 import toast from "react-hot-toast";
 
-import { useAuth } from "../api/useAuth";
+import ShareListForm from "../components/forms/ShareListForm";
 
 interface Props {
 	listPath: string | null;
@@ -22,13 +22,10 @@ const purchaseTimelines = {
 };
 
 export function ManageList({ listPath }: Props) {
-	const { user: currentUser } = useAuth();
-
 	const [itemName, setItemName] = useState("");
 	const [itemNextPurchaseTimeline, setItemNextPurchaseTimeline] = useState(
 		PurchaseTime.soon,
 	);
-	const [emailName, setEmailName] = useState("");
 
 	const handleItemNameTextChange = (e: ChangeEvent<HTMLInputElement>) => {
 		setItemName(e.target.value);
@@ -36,10 +33,6 @@ export function ManageList({ listPath }: Props) {
 
 	const handleNextPurchaseChange = (changed: PurchaseTime) => {
 		setItemNextPurchaseTimeline(changed);
-	};
-
-	const handleEmailNameChange = (e: ChangeEvent<HTMLInputElement>) => {
-		setEmailName(e.target.value);
 	};
 
 	const handleSubmit = async (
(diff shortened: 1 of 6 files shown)


RossaMania:
It was great working with you this week Maha! 

I know we are not supposed to worry about styling, and I love that the acceptance criteria gave us some freedom! Do you think it would be a good idea to add something like a horizontal row (<hr />) between the two forms where that arrow is pointing?

How do you feel about placeholder text? What do you think about putting in "Enter e-mail address. . ." in the text field?

image

RossaMania:
I'm so glad we got some more practice with React Hot Toast this week!

What do you think about changing the messages? Something like `Successfully invited ${emailName} to your list!` for success? Something like `Oops! Failed to invite ${emailName} to your list!. Please try again!` for error?

The important parts we got done are:
A. The toasts are working!
B. The variable shows up in string interpolation.

src/views/ManageList.tsx
```diff
@@ -153,6 +199,24 @@ export function ManageList({ listPath }: Props) {
 						</button>
 					</form>
 					<Toaster />
+					<form onSubmit={(e) => handleInvite(e, listPath)}>
+						<label htmlFor="receipient-email">
```
bbland1, line 203: typo: there's an extra `e` in `recipient` here!

src/views/ManageList.tsx
```diff
@@ -153,6 +199,24 @@ export function ManageList({ listPath }: Props) {
...
+								required
+								aria-label="Enter the user name to share list"
+								aria-required
+							/>
+						</label>
+						<button type="submit" aria-label="Add item to shopping list">
+							Invite User
+						</button>
+					</form>
+					<Toaster />
```
bbland1, line 219: note: I don't think there needs to be 2 `<Toaster />` elements in this. It should only need one.

eternalmaha:
> I'm so glad we got some more practice with React Hot Toast this week!
> 
> What do you think about changing the messages? Something like `Successfully invited ${emailName} to your list!` for success? Something like `Oops! Failed to invite ${emailName} to your list. Please try again!` for error?
> 
> The important parts we got done are: A. The toasts are working! B. The variable shows up in string interpolation.

src/views/ManageList.tsx
```diff
@@ -76,6 +85,43 @@ export function ManageList({ listPath }: Props) {
...
 	};
 
+	const handleInvite = async (
+		e: FormEvent<HTMLFormElement>,
+		listPath: string,
+	) => {
+		console.log("Button clicked! Inviting user!");
+		e.preventDefault();
+
+		const trimmedEmailName = validateTrimmedString(emailName);
```
bbland1, line 95: praise: I love seeing this util function `validateTrimmedString` being used again! How cool!

src/views/ManageList.tsx
```diff
@@ -76,6 +85,43 @@ export function ManageList({ listPath }: Props) {
...
+		}
+
+		if (currentUser === null) {
+			toast.error("You are not logged in! Cannot invite.");
+			return;
+		}
+
+		try {
+			await toast.promise(shareList(listPath, currentUser, trimmedEmailName), {
+				loading: "sharing list with existing user",
```
bbland1, line 111: > I'm so glad we got some more practice with React Hot Toast this week!

What do you think about changing the messages? Something like `Successfully invited ${emailName} to your list!` for success? Something like `Oops! Failed to invite ${emailName} to your list. Please try again!` for error?

The important parts we got done are:
A. The toasts are working!
B. The variable shows up in string interpolation. 

I really like changing the messages to be a bit more descriptive! It would allow another way for the user to see like "who did i just invite?".

src/views/ManageList.tsx
```diff
@@ -20,10 +22,13 @@ const purchaseTimelines = {
 };
 
 export function ManageList({ listPath }: Props) {
+	const { user: currentUser } = useAuth();
```
bbland1, line 25: question: Was there a reason to call `useAuth()` here? I know the user is needed on [line 110](https://github.com/the-collab-lab/tcl-77-smart-shopping-list/blob/rc-ma-share-shopping-list/src/views/ManageList.tsx?plain=1#L110) and I'm just curious if it was considered adding the `user` prop to the `<ManageList />` in `App.tsx` and add it to the props in the call here.

src/views/ManageList.tsx
```diff
@@ -153,6 +199,24 @@ export function ManageList({ listPath }: Props) {
 						</button>
 					</form>
 					<Toaster />
+					<form onSubmit={(e) => handleInvite(e, listPath)}>
+						<label htmlFor="receipient-email">
+							<input
```
bbland1, line 204: suggestion: It would be nice to add the label value here. So that is shows it before the input.

```suggestion
<label htmlFor="recipient-email">
    Recipient Email:
    <input
```

bbland1:
I'm not really sure where to put this so I'm just throwing it in a PR comment over all. I noticed when I typed in the email I am signed up with and an email of a not currently existing in the DB user I got a successful message both times. 

Reading over the acceptance criteria

> [ ] If the other user does not exist, the user is shown an error message that explains the problem

I think it should get an error at least if an email is passed of a user that does not currently exist within the application.

I'm not sure if it should send some type of error if the user tries to share the list with themselves acceptance-wise but it may be nice to think about!



https://github.com/user-attachments/assets/45e3d4b6-a6f5-482a-86a9-e0eb49232f7d

eternalmaha:
Wow! I love how we noticed that at the exact same time :) "Async" hahah. You're absolutely right!! I just realized that it does in fact send a "successful notification" regardless of whether the user exists or not (any email was working). After looking at the shareList function that checks whether the invited user exists before sharing the list, the condition was written to "exit" the function early through a return statement if the "invited user" didn't exist. I learned that JavaScript considers a function "successful" if it returns a value or resolves itself without returning an error even if the function never completes. I believe then, the toast promise's error block didn't register the function exiting early as an "error".  So interesting! Thanks for your detailed review @bbland1 ! 

Screenshot 2024-08-29 at 12 21 12 AM
Screenshot 2024-08-29 at 12 42 56 AM

src/views/ManageList.tsx
RossaMania: So cool! Maha and I were wondering that! We figured one form === one toaster element.

src/views/ManageList.tsx
RossaMania: Nothing wrong with answering questions! We can all work together. 😄

src/views/ManageList.tsx
RossaMania: It was not! That's all we could think to do since currentUser had to come from somewhere.

RossaMania:
Great catches you guys! It's not letting me reply to comments directly! Yes. I agree that throwing an actual Error object with a message would help! I feel like the message should say something like "Unable to share list". A message like "Recipient doesn't exist" could cause security concerns. It could be seen as it has just confirmed to somebody that that is not valid user information.

bbland1:
That is awesome @eternalmaha! I saw it and I was like ooo I'm not sure I can describe it well so chose to take the screen recording lol 😅! 

I also think getting the message clear so not to cause concerns is important @RossaMania!

src/views/ManageList.tsx
eternalmaha: good catch!

src/views/ManageList.tsx
eternalmaha: Yeah! We really were curious about that. I thought it was a <Toaster/> per form as well. After looking into it some more, its good practice to "import { Toaster } from 'react-hot-toast' "  into the root of the application , probably in our App.tsx file. By doing that,  we will only need to  " import { toast } from 'react-hot-toast' " into any of our needed files, but we wont need to place any <Toaster/> elements since all our app components will have access to it :)

src/views/ManageList.tsx
eternalmaha: Yes! I like that idea too. I was playing around in the useAuth.tsx file, and I noticed we have access to user.name and can invoke the list owner/user's name into our notifications, but it would be so cool if we could insert the invited user's name instead of their email !!!!

eternalmaha:
@RossaMania I think editing our wording in the notifications sent to the user to be more clear is a great suggestion! I think including that the user is "unable to share list" as you mentioned is important. I think it is also a good idea to include why the list can't be shared so the user is able to troubleshoot, that doesn't cause them unecessary alarm. Perhaps "Unable to share list. Verify user has an active account !" ? 

I'm also down to make some basic styling changes to the appearance as well when we meet tomorrow to review and finalize all the suggested changes :)

RossaMania:
So cool! @eternalmaha just got done meeting!
- We looked at everybody's suggestions so far!
- We corrected spelling errors.
- We changed the wording of toast messages in ManageList components!
- Imported and rendered Toaster component at app level!
- Added placeholder text for that email input, and added some spacing between forms with a hr element.
- Removed all redundant Toaster component renders!

image

src/views/ManageList.tsx
```diff
@@ -76,6 +85,44 @@ export function ManageList({ listPath }: Props) {
...
+
+		try {
+			await toast.promise(shareList(listPath, currentUser, trimmedEmailName), {
+				loading: "sharing list with existing user",
+				success: () => {
+					setEmailName("");
+					return `Successfully invited ${emailName} to your list!`;
+				},
+				error: () => {
+					return `Oops! Failed to invite ${emailName} to your list. Please verify correct email!`;
```
bbland1, line 118: praise: I really like this update to have the email name in the messages!

bbland1 (approved):

zahrafalak:
First off, I want to say that you’ve done a fantastic job with the ManageList component! The way you've structured the core functionality is clear, and your use of react-hot-toast for user feedback adds a nice, polished touch to the user experience. Also, the validation checks you’ve implemented are spot-on, ensuring that only valid data gets processed. 🚀

That said, I have a suggestion that could make the code even more maintainable, and it's one of the things @tannaurus called out for in our code as well, which I think is a great thing to be mindful of when building React apps: 😄

To further enhance the structure, consider breaking the ManageList component down into smaller, focused components like ItemForm, InviteForm, and PurchaseTimelineSelector. This would not only keep the code more organized and easier to manage but would also make these components more reusable across the app, saving time and effort in future development. By separating the item addition and user invitation logic into distinct components, we can improve the overall clarity of the code, making it much easier to test and debug. Additionally, this approach would significantly boost the readability of the code and lay a strong foundation for any future enhancements or updates, ensuring that the code remains clean and maintainable as the project grows. 

Again, excellent work on what you’ve built. The functionality is well-implemented, and the comment above is just a suggestion at this point, and can be implemented as a separate issue on it's own in upcoming future.

src/views/ManageList.tsx
```diff
@@ -76,6 +85,44 @@ export function ManageList({ listPath }: Props) {
...
+
+		if (!trimmedEmailName) {
+			toast.error(
+				"Oops! Email cannot be empty or just spaces. Please enter a valid email.",
+			);
+			return;
+		}
+
+		if (currentUser === null) {
+			toast.error(`You are not logged in! Cannot invite.`);
```
tannaurus, line 105: I think it would be best to not render this component at all instead of notifying the user that they cannot do this operation after they have already attempted to do so

RossaMania:
Yes please! @zahrafalak I agree wholeheartedly! At this point, I feel like our app is getting to a nice size, and I was thinking about bringing up refactoring some elements into separate components this week.
- What nested folders should components be in, in the components folder?
- Are we going to do the arrow functional component export, or continue with the "export function" version?
- Should we wait for the auth routes middleware to be done?
- I can't think of the other questions I was thinking of, the other day.

src/views/ManageList.tsx
RossaMania: Yes please! We were getting an odd message (that I can't remember right now), and that null-check solved the issue.
BUT, thinking about it now, we could just do a ternary operator like,

{currentUser && //code here}

right?

src/views/ManageList.tsx
tannaurus: I do too! Except I don't think it needs to be 😄 because the input element has the `email` type, it isn't possible to get this component into a state where `email` has any whitespace, the input trims it for you already:

https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email)

src/views/ManageList.tsx
```diff
@@ -76,6 +85,44 @@ export function ManageList({ listPath }: Props) {
...
+	const handleInvite = async (
+		e: FormEvent<HTMLFormElement>,
+		listPath: string,
+	) => {
+		console.log("Button clicked! Inviting user!");
+		e.preventDefault();
+
+		const trimmedEmailName = validateTrimmedString(emailName);
+
+		if (!trimmedEmailName) {
```
tannaurus, line 97: Similar to my note above, we can remove this check: it's not possible to hit it.

src/views/ManageList.tsx
```diff
@@ -76,6 +85,44 @@ export function ManageList({ listPath }: Props) {
 		}
 	};
 
+	const handleInvite = async (
```
tannaurus, line 88: Could all of this new logic be its own component that is rendered here instead? There's a problem developing in this file where 1. the component is doing too many things (re: [this other comment I left on PR #23](https://github.com/the-collab-lab/tcl-77-smart-shopping-list/pull/23#discussion_r1730064679)) and 2. now there's a function in this component called `handleSubmit` along side two different forms, making it hard to understand what that function is doing unless you read through its definition. This can be solved quickly by pushing this logic into its component and rendering it here instead 😄

src/views/ManageList.tsx
tannaurus: Yup, that's correct! @bbland1 has been working to make this part of the codebase much more straight forward, so this shouldn't be an issue for much longer 😄 https://github.com/the-collab-lab/tcl-77-smart-shopping-list/pull/27/files

src/views/ManageList.tsx
tannaurus: and for what it's worth, I didn't know this either 😄 I knew `email` wouldn't let [an empty string through](https://github.com/the-collab-lab/tcl-77-smart-shopping-list/pull/28/files#r1739852790) (the benefit of using `email` is that you don't have to worry about figuring out if the string is an email, which can't be an empty string). But I discovered through testing that it was _also_ stripping the white space, which is very good to know!

src/views/ManageList.tsx
RossaMania: Yes please! I have an idea of what I want to do! I can work on this!

RossaMania: Here are some screenshots of the component for your viewing pleasure!
image
image

src/views/ManageList.tsx
RossaMania: image
Perfect! Just make it as clean as possible to save on extra refactoring and feelings later!

eternalmaha:
Thank you everyone for the great input and conversation! My recent last commit: 
-I had refactored the form submission to only render if currentUser is logged in. Though after some thought, I realized it is wise to leave the form submission displaying to all users (logged in or not) as it is a crucial feature of the application. 

Instead I suggest it we implement a strict type assertion in the shareList functions' currentUser so shareList only expects a User, and never a null. 

Since currentUser could be User | null, we can set currentUser as User after importing the User object from firebase. This way, TypeStrict only expects a logged in user. I tested it, and it worked. Next thing would be to add an alert to the user if they try to share a list, and are not logged in. 

The other option would be to do a conditional check before the shareList function of user being logged in or not. Feel free to check any gaps in my logic! If everyone is on board, I'll go ahead and commit the change. 

Screenshot 2024-08-31 at 8 10 09 PM

https://github.com/user-attachments/assets/4fd35c33-0bea-415e-82e0-fb7cd5c94caf

src/views/ManageList.tsx
eternalmaha: WooooHoooo!!!!!

PR #29, opened 2024-08-29
Issue 7: Adding list filtering

### LINKED ISSUE #7: 7. As a user, I want to filter my shopping list to make it easier to locate an item in the list
## Summary
Allowing users to search through their list will help them cut down on clutter and focus on items they’re looking for.

## Acceptance criteria
- [x] A form is added to the top of the `List` view, above the shopping list
- [x] The form includes the following elements
	- A text field (with semantic `<label>`!) which narrows down the list as the user types
	- When there’s text in the field, some kind of button (e.g., an X) to clear the field. When the field is cleared, the list is reset to its unfiltered state.

## Notes
- You will know your `<label>` is correctly implemented if you click on the label and keyboard focus moves to the related input. Refer to [the MDN docs on `<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) for more info.
- The filter text should match *any* part of the item name (e.g., it should _not_ only match from the start of the string)

bbland1:

## Description

Allowing users to search through their list will help them cut down on clutter and focus on items they’re looking for.

## Related Issue

Closes #7 

## Acceptance Criteria

- [x] A form is added to the top of the `List` view, above the shopping list
- [x] The form includes the following elements
	- A text field (with semantic `<label>`!) which narrows down the list as the user types
	- When there’s text in the field, some kind of button (e.g., an X) to clear the field. When the field is cleared, the list is reset to its unfiltered state.

## Type of Changes

|  ✓    | 💯 Enhancement |
|  ✓    | ✨ New Feature |

### Before

Screenshot 2024-08-28 at 10 22 40 PM

### After

https://github.com/user-attachments/assets/46320916-29ac-44ce-8e57-a7d9051b1d6d

- Visit the [preview URL](https://tcl-77-smart-shopping-list--pr29-fz-bb-filter-shoppin-jmib3sm3.web.app/) for this PR
- Sign in to the application
- Click a list that you want view on homepage
- Navigate to list view
- Type into the the search bar at the top of the list
- Hit the `x` in the box when ready to clear the search

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
 		"eslint-plugin-react": "^7.35.0",
 		"husky": "^9.1.4",
 		"jsdom": "^24.1.1",
-		"lint-staged": "^15.2.8",
+		"lint-staged": "^15.2.9",
 		"prettier": "^3.3.3",
 		"typescript-eslint": "^8.2.0",
 		"vite": "^5.3.5",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,3 @@
-import React from "react";
-
 import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
 
 import { Home, Layout, List, ManageList } from "./views";
--- a/src/components/FilterList.tsx
+++ b/src/components/FilterList.tsx
@@ -11,20 +11,18 @@ export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
 	};
 
 	return (
-		<>
-			<form>
-				<label htmlFor="filterList">
-					Filter List:
-					<input
-						type="search"
-						onChange={handleChange}
-						value={searchTerm}
-						id="filterList"
-						aria-label="Filter items in the list"
-						placeholder="Search items..."
-					/>
-				</label>
-			</form>
-		</>
+		<form>
+			<label htmlFor="filterList">
+				Filter List:
+				<input
+					type="search"
+					onChange={handleChange}
+					value={searchTerm}
+					id="filterList"
+					aria-label="Filter items in the list"
+					placeholder="Search items..."
+				/>
+			</label>
+		</form>
 	);
 }
--- a/src/components/FilterList.tsx
+++ b/src/components/FilterList.tsx
@@ -1,5 +1,4 @@
 import { ChangeEvent } from "react";
-import { ListItem } from "../api";
 
 interface FilterListProps {
 	searchTerm: string;
@@ -10,6 +9,7 @@ export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
 	const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
 		setSearchTerm(e.target.value);
 	};
+
 	return (
 		<>
 			<form>
--- a/src/components/FilterList.tsx
+++ b/src/components/FilterList.tsx
@@ -16,7 +16,7 @@ export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
 				<label htmlFor="filterList">
 					Filter List:
 					<input
-						type="text"
+						type="search"
 						onChange={handleChange}
 						value={searchTerm}
 						id="filterList"
--- a/src/components/FilterList.tsx
+++ b/src/components/FilterList.tsx
@@ -1,0 +1,32 @@
+import { ChangeEvent } from "react";
+import { ListItem } from "../api";
+
+interface FilterListProps {
+	searchTerm: string;
+	// setSearchTerm: React.Dispatch<React.SetStateAction<string>>
+	setSearchTerm: (term: string) => void;
+}
+
+export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
+	const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+		setSearchTerm(e.target.value);
+	};
+	return (
+		<>
+			<form>
+				<label htmlFor="filterList">
+					{" "}
+					Filter List:
+					<input
+						type="text"
+						onChange={handleChange}
+						value={searchTerm}
+						id="filerList"
+						aria-label="Filter Shopping List"
+						placeholder="Search items..."
+					/>
+				</label>
+			</form>
+		</>
+	);
+}
--- a/src/components/FilterList.tsx
+++ b/src/components/FilterList.tsx
@@ -3,7 +3,6 @@ import { ListItem } from "../api";
 
 interface FilterListProps {
 	searchTerm: string;
-	// setSearchTerm: React.Dispatch<React.SetStateAction<string>>
 	setSearchTerm: (term: string) => void;
 }
 
@@ -15,13 +14,12 @@ export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
 		<>
 			<form>
 				<label htmlFor="filterList">
-					{" "}
 					Filter List:
 					<input
 						type="text"
 						onChange={handleChange}
 						value={searchTerm}
-						id="filerList"
+						id="filterList"
 						aria-label="Filter Shopping List"
 						placeholder="Search items..."
 					/>
--- a/src/components/FilterList.tsx
+++ b/src/components/FilterList.tsx
@@ -20,7 +20,7 @@ export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
 						onChange={handleChange}
 						value={searchTerm}
 						id="filterList"
-						aria-label="Filter Shopping List"
+						aria-label="Filter items in the list"
 						placeholder="Search items..."
 					/>
 				</label>
--- a/src/components/FilterListInput.tsx
+++ b/src/components/FilterListInput.tsx
@@ -5,7 +5,10 @@ interface FilterListProps {
 	setSearchTerm: (term: string) => void;
 }
 
-export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
+export function FilterListInput({
+	searchTerm,
+	setSearchTerm,
+}: FilterListProps) {
 	const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
 		setSearchTerm(e.target.value);
 	};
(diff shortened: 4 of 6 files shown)


src/components/FilterList.tsx
```diff
@@ -0,0 +1,30 @@
+import { ChangeEvent } from "react";
+
+interface FilterListProps {
+	searchTerm: string;
+	setSearchTerm: (term: string) => void;
+}
+
+export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
+	const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+		setSearchTerm(e.target.value);
```
eternalmaha, line 10: I noticed when I console.log(searchTerm) after the state update, the searchTerm is missing a letter. I think using a useEffect to obtain the latest updated state after the user inputs a search term would be helpful  to get around React's asynchronous nature of updating states and always maintain the latest state value immediately!

useEffect(() => {
        console.log('Updated searchTerm:', searchTerm);
    }, [searchTerm]);

src/components/FilterList.tsx
```diff
@@ -0,0 +1,30 @@
...
+	};
+
+	return (
+		<>
```
eternalmaha, line 14: Minor! I noticed the use of the fragments <> </> is not needed since there is only one higher level parent element , the form element. :)

package.json
```diff
@@ -31,7 +31,7 @@
 		"eslint-plugin-react": "^7.35.0",
 		"husky": "^9.1.4",
 		"jsdom": "^24.1.1",
-		"lint-staged": "^15.2.8",
+		"lint-staged": "^15.2.9",
```
tannaurus, line 34: I think this is fine to leave in this PR, but generally speaking its best to try to avoid upgrades to dependencies unless they're strictly related to the specific task at hand 😄

src/views/List.tsx
```diff
@@ -1,21 +1,35 @@
...
 
 interface Props {
 	data: ListItem[];
 }
 
 export function List({ data }: Props) {
-	const hasItem = data.length !== 0;
+	const [searchTerm, setSearchTerm] = useState<string>("");
+
+	const filteredData = data.filter((item) =>
```
tannaurus, line 13: We should wrap this in a `useMemo`, which will cache a calculation between re-renders. That way if this component gets re-rendered but `searchTerm` hasn't changed between renders, we don't have to re-filter to list. It can be a big win for performance, especially as this list gets longer

https://react.dev/reference/react/useMemo

src/views/List.tsx
```diff
@@ -1,21 +1,35 @@
...
+		item.name.toLowerCase().includes(searchTerm.toLowerCase()),
```
tannaurus, line 14: Great use of `toLowerCase`!

src/components/FilterList.tsx
```diff
@@ -0,0 +1,30 @@
+import { ChangeEvent } from "react";
+
+interface FilterListProps {
+	searchTerm: string;
+	setSearchTerm: (term: string) => void;
+}
+
+export function FilterList({ searchTerm, setSearchTerm }: FilterListProps) {
```
tannaurus, line 8: Might be best to rename this to something akin to `FilterListInput`; the current name suggests it itself is the list 😄

src/components/FilterList.tsx
tannaurus: Good eye!

src/views/List.tsx
```diff
@@ -1,21 +1,35 @@
...
+
+	const hasItem = filteredData.length !== 0;
+
 	return (
 		<>
 			<p>
 				Hello from the <code>/list</code> page!
 			</p>
+
+			<FilterListComponent
```
tannaurus, line 25: Should we be rendering this if `data` has no length?

src/views/List.tsx
```diff
@@ -1,21 +1,35 @@
...
 
 export function List({ data }: Props) {
-	const hasItem = data.length !== 0;
+	const [searchTerm, setSearchTerm] = useState<string>("");
+
+	const filteredData = data.filter((item) =>
+		item.name.toLowerCase().includes(searchTerm.toLowerCase()),
+	);
+
+	const hasItem = filteredData.length !== 0;
```
tannaurus, line 17: This I think is the somewhat dangerous part of using variables like this: they can be misleading. This is _really_ `hasFilteredItems`, but even that is hard to mentally grasp onto when we also want to conditionally render `FilterListComponent` if there are any items (regardless of filter). Imo I think it would be easier to understand what this component is doing if we remove the conditional render on line 31 and let `filteredData.map` return an empty array when there's nothing in `filteredData`; it will have the same effect without the indirection introduced by `hasItem` and make conditionally rendered `FilterListComponent` under separate but similar conditions easier to understand what makes them different from each other

src/views/List.tsx
```diff
@@ -1,21 +1,35 @@
...
+import { useState } from "react";
+import { ListItem as ListItemComponent } from "../components/ListItem";
+import { FilterList as FilterListComponent } from "../components/FilterList";
+import { ListItem } from "../api";
 
 interface Props {
 	data: ListItem[];
 }
 
 export function List({ data }: Props) {
```
tannaurus, line 10: One thing I love to do in situations like this is rename props on the fly based on the context of how the component is using them. Sometimes its helpful to call the prop something simple like `data`, but inside the component that "data" has a different meaning: it's really `unfilteredListItems`. You can change the name of the variable when you deconstruct the `Props` object like this `List({ data: unfilteredListData })` which can bring new contextual meaning to your references of `data` without changing what the function signature looks like to the parent component: it just thinks of this as "data" and "you" get to think of it as what it rightfully is: unfiltered list data

tannaurus (approved):

tannaurus:
Left some tiny thoughts but approved! Great work!

src/components/FilterList.tsx
tannaurus: I'm having a hard time reproducing this on my end. React _should_ be handling this for us already 👀

src/App.tsx
RossaMania, line 1: Nice catch removing the React import! Since we're using that ".tsx" file format, the React import is redundant!

src/views/List.tsx
```diff
@@ -1,21 +1,35 @@
...
+
+			<FilterListComponent
+				searchTerm={searchTerm}
+				setSearchTerm={setSearchTerm}
+			/>
+
 			<ul>
 				{hasItem &&
-					data.map((item) => (
+					filteredData.map((item) => (
```
RossaMania, line 32: Praise: All of this functionality looks clean!
Nitpick: Why not call this variable filteredListData?

src/components/FilterList.tsx
bbland1: Very good catch thank you!

src/views/List.tsx
bbland1: I love this and the ability to give the code more context!

src/views/List.tsx
bbland1: This is a really good point, and ties in with your comment about renaming the prop on the fly, and @RossaMania comment on updating the name of the `filterdData` as well. I think all those changes make the flow of the code easier to read!

src/views/List.tsx
bbland1: Good point, what is there to filter if there is no data. It's not really visually needed.

src/components/FilterList.tsx
eternalmaha: Uploading Screen Recording 2024-09-01 at 12.30.47 AM.mov…

eternalmaha: I dont think its really an issue, but its more an observation

https://github.com/user-attachments/assets/c15e31e6-883a-44e0-be00-d2ccedef4a27

bbland1: I'm also not sure I can reproduce this on my end 🤔

bbland1: Thanks for the screen record! I did get too when I have been trying to recreate it. I didn't think it was behind in state, which maybe is on my part. I thought it was working as react intended, because the `console.log(searchTerm)` will log the state before the the update takes effect, because the function has to complete before they match, but to my understanding it does filter based on the full input.

src/views/List.tsx
eternalmaha, line 1: This was such a thoughtful addition!!!! big praise!! I like that you added this because it keeps the user experience in mind.

src/components/FilterListInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function FilterListInput({
+	searchTerm,
+	setSearchTerm,
+}: FilterListProps) {
+	const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+		setSearchTerm(e.target.value);
+	};
+
```
eternalmaha, line 15: `nitpick` x 2. Minor change. What do you guys think about adding an onSubmit() to the form to prevent the default refresh in case a user hits enter? I tried clicking enter now after searching for a list item, and the whole page refreshes since its a form element. In the event a user clicks enter, handling the event to remain on the page could be beneficial for a seamless user experience? 
Screenshot 2024-09-01 at 12 59 26 AM

src/views/List.tsx
zahrafalak: That is a great call out. I had not used that hook before, so this gave me a chance to have a go at it.

src/components/FilterList.tsx
eternalmaha: Ahhh gotcha!! That makes sense then.  Natural React behavior. And yes! It is doing the thing its supposed to do exactly: filter the list every single time!!! Great job on the refactoring changes by the way.  I thought I'd bring it up in case I was missing something..

eternalmaha (approved):
Great Job Yall!!! I really enjoyed watching the issue expand and evolve as the days went on!! Seems like you guys experimented with a lot

src/components/FilterListInput.tsx
zahrafalak: Great catch!

PR #30, opened 2024-09-01
moving our auth/unauth routes into new auth/unauth sub directories

bbland1:

## Description

With the mention of moving some files in the views of in PR #27. I added the `authenticaed` and `unauthenticated` subdirectories in views and just moved `ManageList`, `List` and `PageNotFound` in to their proper directories.

## Related Issue

There is no current issue.

## Type of Changes

- There shouldn't be any functional changes, but please check to make sure it works on yalls computers too! 😅


bbland1:
Continuing on a comment that @tannaurus made on PR #27 about organizing the code a bit more as the react app gets bigger. After merging the #27 I made this PR to add the `authenticated` and `unauthenticated` subdirectories within `views` and moved the few files that fall into those (updating imports where needed). 

I wanted to pose the question if there were any additional subdirectories everyone would want to add to `components` it is starting to get bigger and will as we add functionality so I figured while we are doing a move only PR if any suggestions it would be a time! 

I know @RossaMania and @eternalmaha made the `forms` subdirectory and I was thinking maybe `CreateList` and `FilterListInput` could go in there since they are all forms. Then maybe some other ones but wasn't sure so wanted to make the PR to ask 😄

RossaMania:
How do we feel about an AddItemForm.tsx component? I feel like all the logic for adding items and selecting when to buy, as well as lines 86-158 in views/ManageList.tsx could be in this component.

bbland1:
> How do we feel about an AddItemForm.tsx component? I feel like all the logic for adding items and selecting when to buy, as well as lines 86-158 in views/ManageList.tsx could be in this component.

I am not oppose to that at all, I think that logic in `ManageList` should be its own component. 

What I am not not sure on if adding that file and moving that would fall into a reorganization. I took it as we are moving the files we have to be in better organized sub directories not moving any of the logic as of now.🤔

alex-andria (approved):
Pretty simple refactor! I see this is from the recommendation by Tanner in creating a cleaner directory structure for authenticated and unauthenticated directories from PR #27 like you said.

Still works functionally when testing. If the other Collabies are on the same page and approved this one should be good to go! Thanks for the initiative in getting this PR out @bbland1 🙌 🙌

tannaurus:
> > How do we feel about an AddItemForm.tsx component? I feel like all the logic for adding items and selecting when to buy, as well as lines 86-158 in views/ManageList.tsx could be in this component.
> 
> I am not oppose to that at all, I think that logic in `ManageList` should be its own component.
> 
> What I am not not sure on if adding that file and moving that would fall into a reorganization. I took it as we are moving the files we have to be in better organized sub directories not moving any of the logic as of now.🤔

I like both of these ideas a lot! Ross, your proposed refactored sounds great. I agree with Brianna that we could do that in another PR if someone has the bandwidth to pick it up 😄

tannaurus (approved):

alex-andria:
> > How do we feel about an AddItemForm.tsx component? I feel like all the logic for adding items and selecting when to buy, as well as lines 86-158 in views/ManageList.tsx could be in this component.
> 
> I am not oppose to that at all, I think that logic in `ManageList` should be its own component.
> 
> What I am not not sure on if adding that file and moving that would fall into a reorganization. I took it as we are moving the files we have to be in better organized sub directories not moving any of the logic as of now.🤔

That's a good point! For best practice, I know smaller PR changes are best. However in this case, if you were to make that change and it works functionally I don't see a problem with making that refactor here if that's what makes sense for your group and you all agree. We just have to make sure these changes are pulled before merging the upcoming PRs for this week.

bbland1:
I'm gonna merge this so that. There can be later PRs and the code will be the at the same place for other PRs.

PR #31, opened 2024-09-04
Mark item as purchased & update DB

### LINKED ISSUE #9: 9. As a user, I want to mark an item on my shopping list as purchased so the app can learn how often I buy different items
## Summary
Users need a UI that allows them to mark the their items as purchased, so they can track what on their list they do and do not need to buy.

## Acceptance criteria
- [x] The `ListItem` component renders a checkbox with a semantic `<label>`.
- [x] Checking off the item in the UI also updates the `dateLastPurchased` and `totalPurchases` properties on the corresponding Firestore document
- [x] The item is shown as checked for 24 hours after the purchase is made (i.e. we assume the user does not need to buy the item again for at least 1 day). After 24 hours, the item unchecks itself so the user can buy it again.
- [x] The `updateItem` function in `firebase.js` has been filled out, and sends updates to the firestore database when an item is checked

## Notes:
- While you will need to update multiple parts of the Firestore document for this feature, **do not** worry about `dateNextPurchased` yet. That will be addressed in a future issue.
- You can use the Firestore console to test that this feature is working correctly, e.g., by manipulating the value of `dateLastPurchased`

bbland1:
## Description

This change adds the ability for a user to mark an item on their list as purchased. When marking the item as purchased the database will be updated with the current date as the new `dateLastPurchased` and increment the `totalPurchases` of the item. The list item checks if it has been more than 24 hours since an item was marked as purchased the item will uncheck and be able to be marked as purchased again. If a user tries to mark an item as purchased again in 24 hours it will prevent them from triggering it again.

## Related Issue

Closes #9 

## Acceptance Criteria

- [x] The `ListItem` component renders a checkbox with a semantic `<label>`.
- [x] Checking off the item in the UI also updates the `dateLastPurchased` and `totalPurchases` properties on the corresponding Firestore document
- [x] The item is shown as checked for 24 hours after the purchase is made (i.e. we assume the user does not need to buy the item again for at least 1 day). After 24 hours, the item unchecks itself so the user can buy it again.
- [x] The `updateItem` function in `firebase.js` has been filled out, and sends updates to the firestore database when an item is checked

## Type of Changes

|  ✓    | 💯 Enhancement |
|  ✓    | ✨ New Feature |

### Before

Screenshot 2024-09-05 at 11 46 07 AM

### After

Screenshot 2024-09-06 at 1 59 11 PM

https://github.com/user-attachments/assets/67823bd4-405a-4a11-9fe0-9f1ff97d6eaf

- From your terminal, pull down this branch with `git pull origin bb-ma/mark-item-purchased and check that branch out with git checkout `bb-ma/mark-item-purchased`
- Then `npm ci` to install the newly added dependencies locally and `npm start` to launch the app.
- Click one of your lists that has items in it.
- Click the list navigate link in the nav bar.
- All items in the list should now have a checkbox to the left of them.
- Click either the label or the checkbox to mark an item as purchased.
- Click again and you will get a notification that will say it can't be marked again in a 24 hour period. 
- Open a new tab with the [firebase console](https://console.firebase.google.com/u/0/project/tcl-77-smart-shopping-list/firestore/databases/-default-/data/~2F5mLi0TYfwKVk9BnM3fdRRR9BaYe2~2FCostco) and navigate to your lists (the unique id associated with you outside of the users collection).
- Open the list you just marked items purchased on.
- See that the `dateLastPurchased` and `totalPurchases` has been updated.
- Change the `dateLastPurchased` to be at least 24 hours and 1 minute in the past from the current `dateLastPurchased`.
- Return to the app and notice the checkbox should be unchecked and able to be updated again.


eternalmaha:
I was looking into the setTimout() built in Javascript method that could be useful for UI purposes of unchecking the checkbox. If we needed to update anything in the backend as well, i'm sure it could be useful to for coordinating with the firebase document.

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+		? item.dateLastPurchased.toDate()
+		: item.dateLastPurchased;
+
+	useEffect(() => {
+		if (purchaseDate) {
+			setChecked(!is24HoursLater(purchaseDate));
+		} else {
+			setChecked(false);
+		}
+	}, [purchaseDate]);
```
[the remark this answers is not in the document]
eternalmaha: great catch!!! It would be better to track the raw value of the dateLastPurchased to avoid any issues

src/utils/dates.ts
```diff
@@ -10,3 +10,13 @@ const ONE_DAY_IN_MILLISECONDS = 86400000;
...
 }
+
+export function is24HoursLater(purchaseDate: Date): boolean {
+	const currentTime = new Date();
+
+	const timeElapsedInMilliseconds = Math.abs(
+		currentTime.getTime() - purchaseDate.getTime(),
+	);
+
+	return timeElapsedInMilliseconds >= ONE_DAY_IN_MILLISECONDS;
```
RossaMania, line 21: I like that this is it's own function in it's own file.

Bonus: it's in a file that's already in a utils folder! So clean!

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+	useEffect(() => {
+		if (purchaseDate) {
+			setChecked(!is24HoursLater(purchaseDate));
+		} else {
+			setChecked(false);
+		}
+	}, [item.dateLastPurchased]);
+
+	const handleCheckChange = async (e: React.ChangeEvent<HTMLInputElement>) => {
+		const isChecked = e.target.checked;
```
RossaMania, line 28: Nice use of e.target.checked, instead of e.target.value!

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+	listPath: string | null;
+}
+
+export function ListItemCheckBox({ item, listPath }: Props) {
+	const [checked, setChecked] = useState(false);
+
+	const purchaseDate = item.dateLastPurchased
+		? item.dateLastPurchased.toDate()
+		: item.dateLastPurchased;
+
```
RossaMania, line 18: That ternary operator implementation is smooooth!

RossaMania (approved):
I like this! Everything is divided up into functions or functional  components, and even different files/folders.

src/api/firebase.ts
```diff
@@ -245,12 +245,15 @@ export async function addItem(
...
-export async function updateItem() {
-	/**
-	 * TODO: Fill this out so that it uses the correct Firestore function
-	 * to update an existing item. You'll need to figure out what arguments
-	 * this function must accept!
-	 */
+export async function updateItem(listPath: string, item: ListItem) {
+	const itemDocRef = doc(db, listPath, "items", item.id);
+
+	const updates: Partial<ListItem> = {};
```
tannaurus, line 251: Very nice use of `Partial`! This is comes from a family of types in Typescript know as Utility Types. They're one of my favorite parts about the language. Instead of using `Partial` here, which allows for anything, I think this would be a bit more clear

```
const updates: Pick<ListItem, "totalPurchases" | "dateLastPurchased"> = {
 totalPurchases: item.totalPurchases + 1,
 dateLastPurchased: Timestamp.fromDate(new Date())
}
```

Instead of setting the fields later, we can set them directly. Doing that allows us to have a stronger type. Here I'm using a different Utility Type, `Pick`, to select specific fields from the `ListItem` interface I want to be enforced in this type.

By using partial, other code can't tell which fields are actually meant to be there. `totalPurchases` and `dateLastPurchased` could be `null`, and `dateNextPurchased` might also be there. Less applicable for this use case, but always good to be explicit where you can 😄

tannaurus: https://www.typescriptlang.org/docs/handbook/utility-types.html

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
-type Props = Pick<api.ListItem, 'name'>;
+interface Props {
+	item: ListItem;
+	listPath: string | null;
+}
+
+export function ListItemCheckBox({ item, listPath }: Props) {
+	const [checked, setChecked] = useState(false);
+
+	const purchaseDate = item.dateLastPurchased
```
tannaurus, line 15: Probably best to stick with the nullish coalescing operator here. It more explicitly communicates to the reader what you're handling here, and says more with less code 🙂  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+		if (!isChecked && purchaseDate && !is24HoursLater(purchaseDate)) {
+			toast.error(
+				`${item.name} has already been marked as purchased in the last 24 hours.`,
+			);
+			return;
+		}
+
+		setChecked(isChecked);
+
+		if (!listPath) {
```
tannaurus, line 39: Going back to the idea of "making impossible states impossible", this is something that shouldn't ever be undefined here, but we're forced by the type system to contend with it being undefined. That's a "code smell" that logic elsewhere could be improved so we don't need to be handling this. Out of scope for this PR, but something to keep in mind 🙂

src/utils/dates.ts
```diff
@@ -10,3 +10,13 @@ const ONE_DAY_IN_MILLISECONDS = 86400000;
 export function getFutureDate(offset: number) {
 	return new Date(Date.now() + offset * ONE_DAY_IN_MILLISECONDS);
 }
+
+export function is24HoursLater(purchaseDate: Date): boolean {
+	const currentTime = new Date();
+
+	const timeElapsedInMilliseconds = Math.abs(
```
tannaurus, line 17: I think this math is a bit incorrect. Using `Math.abs` here allows this to evaluate "24 hours ago" the same as "24 for hours from now"

src/utils/dates.ts
```diff
@@ -10,3 +10,13 @@ const ONE_DAY_IN_MILLISECONDS = 86400000;
 export function getFutureDate(offset: number) {
 	return new Date(Date.now() + offset * ONE_DAY_IN_MILLISECONDS);
 }
+
+export function is24HoursLater(purchaseDate: Date): boolean {
```
tannaurus, line 14: This could benefit from a more descriptive name. `is24HoursLater` suggests this will only return `true` if it's exactly 24 hours later. `longerThan24HoursFromNow` (or something like that)?

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+			setChecked(false);
+		}
+	}, [item.dateLastPurchased]);
+
+	const handleCheckChange = async (e: React.ChangeEvent<HTMLInputElement>) => {
+		const isChecked = e.target.checked;
+
+		if (!isChecked && purchaseDate && !is24HoursLater(purchaseDate)) {
+			toast.error(
+				`${item.name} has already been marked as purchased in the last 24 hours.`,
```
tannaurus, line 32: I wonder if this is really an error state 🤔 are we saying we want to physically prevent someone from buying things twice? I suppose it could be, but in a real world scenario it would be nice to see an override here. Like, "are you sure?" (that's 100% out of scope here, just food for thought)

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+export function ListItemCheckBox({ item, listPath }: Props) {
+	const [checked, setChecked] = useState(false);
+
+	const purchaseDate = item.dateLastPurchased
+		? item.dateLastPurchased.toDate()
+		: item.dateLastPurchased;
+
+	useEffect(() => {
+		if (purchaseDate) {
+			setChecked(!is24HoursLater(purchaseDate));
```
tannaurus, line 21: I really like how `handleCheckChange` is returning early at the bottom of its conditions. Could we do that here too?

```
if (purchaseDate) {
  setChecked(...)
  return;
}

setChecked(false);

File: src/components/ListItem.tsx [hunk 12]
```diff
@@ -1,8 +1,70 @@
...
+		const isChecked = e.target.checked;
+
+		if (!isChecked && purchaseDate && !is24HoursLater(purchaseDate)) {
+			toast.error(
+				`${item.name} has already been marked as purchased in the last 24 hours.`,
+			);
+			return;
+		}
+
+		setChecked(isChecked);
```
tannaurus, line 37: We probably don't want to be setting this yet, it makes error handling harder. For instance, if we ever return an error in the `if (!listPath)` condition, this is left checked despite the database not seeing an update. The user is left misinformed: the ui told them there was an error, but it also appears to have worked. It also would allow us to remove the set we make on line 48

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+			setChecked(!is24HoursLater(purchaseDate));
+		} else {
+			setChecked(false);
+		}
+	}, [item.dateLastPurchased]);
+
+	const handleCheckChange = async (e: React.ChangeEvent<HTMLInputElement>) => {
+		const isChecked = e.target.checked;
+
+		if (!isChecked && purchaseDate && !is24HoursLater(purchaseDate)) {
```
tannaurus, line 30: There's a bit of a "code smell" here that I think is worth thinking about. The idea "state modeling" and "the flow of data" are thoughts I tend to obsess a little about when designing React components. This is an excellent example of why I feel so strongly about the value of that. 

Take a close look at at this condition: how quickly can you figure out what its value is when reading this component, especially as you think about how the inputs could be modifying the value of `isChecked`? I'd argue it takes a bit too long to get to the bottom of that, which is the code smell. `isChecked` could be false because of the `setChecked` on line 23, but it also shares a very close relationship with `purchaseDate`, which here is _also_ being checked.. but does it to be? Wouldn't `isChecked` already be true if `purchaseDate && !is24HoursLater(purchaseDate)` evaluates to `true` because of the check on line 20? I believe this is the same as saying `if(!isChecked && !isChecked)`

Deriving state of props, like we're doing with `isChecked`, can be helpful, but if at any point that state you derived wasn't enough and you start deriving other state on the fly, things can get both very difficult to understand and potentially incorrect. I've got lots more thoughts here, but it's getting late 😴 we can talk sometime soon about all that if you'd like! For now, I think 2/3 of this condition is redundant.

src/utils/dates.ts
```diff
@@ -10,3 +10,13 @@ const ONE_DAY_IN_MILLISECONDS = 86400000;
...
+	const currentTime = new Date();
+
+	const timeElapsedInMilliseconds = Math.abs(
+		currentTime.getTime() - purchaseDate.getTime(),
+	);
+
+	return timeElapsedInMilliseconds >= ONE_DAY_IN_MILLISECONDS;
+}
```
zahrafalak, line 22: Praise: The use of is24HoursLater to control the checkbox state for 24 hours after purchase is clear and intuitive. It directly addresses the requirements.

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+			<label htmlFor={`checkbox-${item.id}`}>
+				<input
+					type="checkbox"
+					id={`checkbox-${item.id}`}
+					aria-label={`Mark ${item.name} as purchased.`}
+					value={item.id}
+					checked={checked}
+					onChange={handleCheckChange}
+					aria-checked={checked}
+				/>
```
zahrafalak, line 65: Suggestion: To improve user experience, the checkbox could be disabled while the item is within the 24-hour period. This would give a clear visual indication that users can't uncheck the item before 24 hours have passed.
You can add it as a property in the input label
`disabled={!is24HoursLater(purchaseDate)} // Disable if 24 hours haven't passed`

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+		<div className="ListItem">
+			<label htmlFor={`checkbox-${item.id}`}>
+				<input
+					type="checkbox"
+					id={`checkbox-${item.id}`}
+					aria-label={`Mark ${item.name} as purchased.`}
+					value={item.id}
+					checked={checked}
+					onChange={handleCheckChange}
+					aria-checked={checked}
```
zahrafalak, line 64: Praise: Great accessibility considerations with the aria-label and label for the checkbox.

src/components/ListItem.tsx
```diff
@@ -1,8 +1,70 @@
...
+		: item.dateLastPurchased;
+
+	useEffect(() => {
+		if (purchaseDate) {
+			setChecked(!is24HoursLater(purchaseDate));
+		} else {
+			setChecked(false);
+		}
+	}, [item.dateLastPurchased]);
+
```
zahrafalak, line 26: Suggestion: The useEffect hook re-calculates the checked state every time item.dateLastPurchased changes. This is fine, but it could be simplified slightly:
```
useEffect(() => {
  setChecked(purchaseDate ? !is24HoursLater(purchaseDate) : false);
}, [purchaseDate]);
```
This version uses `purchaseDate` directly as a dependency in the useEffect instead of `item.dateLastPurchased`, and also reduces the cognitive load inside the hook.

src/api/firebase.ts
```diff
@@ -245,12 +245,15 @@ export async function addItem(
...
+
+	updates.totalPurchases = item.totalPurchases + 1;
+	updates.dateLastPurchased = Timestamp.fromDate(new Date());
+
+	await updateDoc(itemDocRef, updates);
```
zahrafalak, line 256: Suggestion: We can maybe wrap this in a `try...catch` block to handle potential errors in case the `updateDoc` call fails?

src/api/firebase.ts
bbland1: I really like this! I had thought about it being just some of the types we need so when figuring out how to utilize the `ListItem` type again the Partial made sense and most I saw used like an `Omit` I think it was but `Pick` makes so much sense! The setting of the  updates directly is such a good point as well!

src/api/firebase.ts
bbland1: Good catch!

src/components/ListItem.tsx
bbland1: We had talked about more of a prompt that says "are you sure" of some type that will allow it to possibly still buy it twice but it did seem out of scope for this PR and we didn't want it to keep updating the DB when that may not be the intent. So I did think that having an error get thrown would something that is maybe a temporary solution.

src/components/ListItem.tsx
bbland1: Since javascript date objects are mutable I had suggested that the depenency array use the "source of truth" `item.dateLastPurchased`, it would make sure any change in the `useEffect` only happen if there is a change in the `item.dateLastPurchased` keeping it reliable and without the concern of if `purchaseDate` got mutated at some point.

I could be not thinking about the direction quite right. Not sure if @tannaurus, @alex-andria or @adidalal would be able to explain it a bit more.

src/utils/dates.ts
bbland1: You are absolutely right! I always knew simple math would haunt me 🤦🏽‍♀️😂

src/components/ListItem.tsx
tannaurus: Frankly it's been too longer since I've written JS to know confidentially that will work, but I think it will 🙂

tannaurus: Also, tagged unions are by far the best feature in Typescript. I live and breath them for component state modeling

tannaurus: I just realized I used `.await` 🤦 you'll have to translate that for me to valid JS. Whoops!

src/api/firebase.ts
tannaurus: Yes, great catch @zahrafalak !

src/utils/dates.ts
tannaurus: This is the kind of thing unit tests are very _very_ good at. Simple math haunts us all 🙂

src/components/ListItem.tsx
tannaurus: That makes sense 👍 I think @zahrafalak 's suggestion here could be a better way of handling this 😄 If they can't do this action without seeing an error, it's probably best to just not let them do it at all https://github.com/the-collab-lab/tcl-77-smart-shopping-list/pull/31/files#r1747943276

tannaurus (approved):

src/components/ListItem.tsx
bbland1: purchaseDate


I was working on the change of using the nullish coalescing, and when I made the change I started getting a type error with it because in the function `!moreThan24HoursPassed(purchaseDate));` the type of Timestamp isn't Data. 
```
Argument of type 'Timestamp | Date' is not assignable to parameter of type 'Date'.
  Type 'Timestamp' is missing the following properties from type 'Date': toDateString, toTimeString, toLocaleDateString, toLocaleTimeString, and 37 more.ts(2345)
```

I don't necessarily think the 3rd option is the full go to either but I was curious if it would work. I guess it also could just be the `item.dateLastPurchased?.toDate() ?? null` but that feels kind of against the whole point of the nullish coalescing.

src/components/ListItem.tsx
bbland1: I'd love to talk about this more and just get more of your insight!

src/components/ListItem.tsx
bbland1: @tannaurus is this for the useEffect for the handling of the unchecking at 24 hours? Like is having the no use state would handle this without the useEffect?

src/api/firebase.ts
eternalmaha: What a handy and flexible tool!!! My prior understanding of 'Pick' was being able to create a type from another Type model as a means to specify the kind of values your properties will have ("string", "number", etc), but the option to even go further and write blocks of logic as you set the type..is next level.

src/components/ListItem.tsx
alex-andria: I think it would be cool to implement tagged unions in this way since we're going the TS route! I do agree that using the `dateLastPurchased` would be the most accurate way to pull that info from firebase, which is how I've done it in the past. 

If you decide to go the state checking route, which I think I've found common with regular React apps for TCL in the past, it would be a fun exercise to use tagged unions to differentiate local state and firebase state like Tanner said for our TS app and then on firebase query have the local override (update) the storage. 

We can maybe do this during another mini office hours together and make it a refactor spike card for the next sprint. I'm a bit rusty on my TS but open to exploring this soon! We can always add/enhance our modal rendering later too as a UI card later in the cohort, but at the least have our error catching logic in place in the meantime.

alex-andria (approved):
Great work Brianna and Maha!  🙌 🙌 this definitely was a hefty card that really required you to map out how to connect your local storage with the db in firebase. Really impressed with how you two communicated on Slack and tried different solutions for the acceptance criteria. 

Not much I would change on my end besides tightening up the `setChecked` state and exploring the utility of the TS feature of tagged union types - which you'll definitely see in the workplace that uses TS. Also appreciate y'all updating the README, nice touch!

src/components/ListItem.tsx
eternalmaha: Great insight @bbland1 and really helpful article @tannaurus !! I actually think the 3rd option is the best one for our specific use case! The case of item.dateLastPurchased being "null" or "undefined" is exactly what we want to capture so I think we can omit the null coalescing operator that would otherwise run the "right hand" evaluation if the "left-hand" side was null?!?  Thats my interpretation of the logic!  Really neat to see how this omission of the null coalescing operator is precisely because what we are clearly checking for ("null" or "undefined" purchase dates!)

src/utils/dates.ts
eternalmaha: This is an interesting take! I had to wrap my head around this one because I initially thought if purchase date is (logically) always in the past, the 24 hour difference should technically always be positive and then Math.abs wouldn't make a difference?! Then I began to think, there could be cases perhaps if purchase date was for some reason, set to the future and the function is24hoursLater wouldn't be checking the correct logic...in any case, I think it is really valuable to capture the direction of the change in time either way!!

src/components/ListItem.tsx
eternalmaha: yessss!!! This was definitely in sync with some of the conversations we were having. Thanks for the validation and provided code! I like this better than the error message we sent telling them "they already purchased it in 24 hours". While the goals of both cases are the same, its less jarring to not see an error message thrown to the user.

src/components/ListItem.tsx
eternalmaha: Hmm!!! I really appreciate this feedback and convo.  I can now see since we already update item.dateLastPurchased anyways within our updateItem call, to save the cognitive load as Falak was mentioning, figuring out how to update the checkbox state within that updateItem call could be really efficient!!

PR #34, opened 2024-09-13
Issue # 10 Prevent user from adding duplicate or empty item to the list

### LINKED ISSUE #10: 10. As a user, I want to be alerted when I’m entering an empty item or an item that's already on my list
## Summary
Users shouldn't be able to add an empty item to their list, or add the same item twice. If they try to do this, we need to show them an error message that explains the problem. This way, we’ll prevent some clutter in their lists.

## Acceptance criteria
- [x] Show an error message if the user tries to submit an empty item
- [x] Show an error message if the user tries to submit a new item that is _identical_ to an existing name. For instance, if the list contains `apples` and the user adds `apples`.
- [x] Show an error message if the user tries to submit a new item that matches an existing name with punctuation and casing normalized. For instance, if the list contains `apples` and the user adds `aPples` or `apples,` or `a pples`.
- [x] The user’s original input is saved in the database

##  Notes
- There are a couple of ways to approach this task. If you’re not sure how to begin, _ask your mentors_ for advice!

eternalmaha:
## Description

Users shouldn't be able to add an empty item to their list, or add the same item twice. If they try to do this, we need to show them an error message that explains the problem. This way, we’ll prevent some clutter in their lists.

## Related Issue

closes #10 

## Acceptance Criteria

- [x]  Show an error message if the user tries to submit an empty item
- [x]  Show an error message if the user tries to submit a new item that is identical to an existing name. For instance, if the list contains apples and the user adds apples.
- [x]  Show an error message if the user tries to submit a new item that matches an existing name with punctuation and casing normalized. For instance, if the list contains apples and the user adds aPples or apples, or a pples.
- [x]  The user’s original input is saved in the database

## Type of Changes

`enhancement` 

## Updates

https://github.com/user-attachments/assets/3061c804-2e60-4d9c-a0f6-db47f800bb59

1. Visit the [preview URL for this PR](https://tcl-77-smart-shopping-list--pr34-ma-fz-duplicate-or-e-7fmtso5p.web.app/)
2. Navigate to ManageList view and add a new Item to the list. 
3. Try to add an empty item, which should show an error message. 
4. Try to add an item that is identical to an item that already exists in the list. This should show an error indicating that the item already exists in the list. 
5. Try to add an item that is similar to an existing item with different casing and punctuation (e.g., if the list already contains `Apples`, try adding `apples` or `aPples` or `a pples`). This should show an error indicating that the item already exists in the list.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -70,7 +70,7 @@ export function App() {
 						/>
 						<Route
 							path="/manage-list"
-							element={<ManageList listPath={listPath} />}
+							element={<ManageList listPath={listPath} data={data || []} />}
 						/>
 					</Route>
 
--- a/src/utils/validateTrimmedString.ts
+++ b/src/utils/validateTrimmedString.ts
@@ -13,9 +13,11 @@ export function validateItemName(
 	}
 
 	//Remove punctuation marks and normalize input
-	const punctuationRegex = /[^a-z0-9]/g;
+	const punctuationRegex = /[^\p{L}]/gu;
 
-	const normalizedInputName = input.toLowerCase().replace(/[^a-z0-9]/g, "");
+	const normalizedInputName = trimmedInput
+		.replace(punctuationRegex, "")
+		.toLowerCase();
 
 	//Create a list of  normalized existing item names
 	const normalizedExistingItemNames = existingItems.map((existingItem) => {
--- a/src/utils/validateTrimmedString.ts
+++ b/src/utils/validateTrimmedString.ts
@@ -13,11 +13,9 @@ export function validateItemName(
 	}
 
 	//Remove punctuation marks and normalize input
-	const punctuationRegex = /[^\p{L}]/gu;
+	const punctuationRegex = /[^a-z0-9]/g;
 
-	const normalizedInputName = trimmedInput
-		.replace(punctuationRegex, "")
-		.toLowerCase();
+	const normalizedInputName = input.toLowerCase().replace(/[^a-z0-9]/g, "");
 
 	//Create a list of  normalized existing item names
 	const normalizedExistingItemNames = existingItems.map((existingItem) => {
--- a/src/utils/validateTrimmedString.ts
+++ b/src/utils/validateTrimmedString.ts
@@ -33,7 +33,7 @@ export function validateItemName(
 
 	//return error if the item already exists
 	if (isDuplicateItem(normalizedInputName)) {
-		return "Item already exists in the list.";
+		return ` ${normalizedInputName} already exists in the list`;
 	}
 
 	// Return null if no errors are found (input is valid)
--- a/src/utils/validateTrimmedString.ts
+++ b/src/utils/validateTrimmedString.ts
@@ -1,10 +1,16 @@
 // makes sure the string passed into the function isn't an empty string
-export function validateTrimmedString(input: string) {
-	const trimmedInput = input.trim();
+export function validateItemString(input: string) {
+	const trimmedInput = input.trim(); //removes leading and trailing whitespaces
 
+	// Condition 1: Check if the input is empty
 	if (trimmedInput.length === 0) {
 		return null;
 	}
 
-	return trimmedInput;
+	const removedPunctuation = /[^\p{L}]/gu;
+
+	const validatedString = trimmedInput.replace(removedPunctuation, "");
+	console.log(validatedString);
+
+	return validatedString;
 }
--- a/src/utils/validateTrimmedString.ts
+++ b/src/utils/validateTrimmedString.ts
@@ -10,7 +10,6 @@ export function validateItemString(input: string) {
 	const removedPunctuation = /[^\p{L}]/gu;
 
 	const validatedString = trimmedInput.replace(removedPunctuation, "");
-	console.log(validatedString);
 
-	return validatedString;
+	return validatedString.length > 0 ? validatedString : null;
 }
--- a/src/utils/validateTrimmedString.ts
+++ b/src/utils/validateTrimmedString.ts
@@ -13,7 +13,7 @@ export function validateItemName(
 	}
 
 	//Remove punctuation marks and normalize input
-	const punctuationRegex = /[^\p{L}\s]/gu;
+	const punctuationRegex = /[^\p{L}]/gu;
 
 	const normalizedInputName = trimmedInput
 		.replace(punctuationRegex, "")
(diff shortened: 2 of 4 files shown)


RossaMania:
PRAISE: I LOVE how this feature uses the power regular expressions or "regex"! I love continuing the use of toasts, rather than using text error messages around the input field. I love that the toast error message is short and clear.

NITPICK: I wish the text field would go empty or go blank after receiving the error toast. Now I get the error message, but still have to delete the offending text from the input, before putting in a new item. At the same time though, the input field not being empty DEFINITELY shows that the duplicate item was not added to the list. 

image

src/utils/validateTrimmedString.ts
```diff
@@ -1,10 +1,41 @@
...
+	};
+
+	//return error if the item already exists
+	if (isDuplicateItem(normalizedInputName)) {
+		return ` ${normalizedInputName} already exists in the list`;
 	}
 
-	return trimmedInput;
+	// Return null if no errors are found (input is valid)
+	return null;
```
RossaMania, line 40: I like that the validateItemName func was added to the validateTrimmedString.tsx file as opposed to making a brand new file--all of the functions in this file have to do with validation! Nice!

RossaMania (approved):
Last PRAISE: I like that there are code comments that explain "the why" throughout this in addition to specific variable names! Great job!

src/utils/validateTrimmedString.ts
```diff
@@ -1,10 +1,41 @@
...
+	// Condition 2: Check if the normalized input matches any existing item
+	const isDuplicateItem = (normalizedInputName: string): boolean => {
+		return normalizedExistingItemNames.some(
+			(item) => item === normalizedInputName,
+		);
+	};
+
+	//return error if the item already exists
+	if (isDuplicateItem(normalizedInputName)) {
+		return ` ${normalizedInputName} already exists in the list`;
```
alex-andria, line 36: Nice job specifying to the user the exact error when duplicate entries are submitted!

alex-andria (approved):
Great job with this PR Falak and Maha! I know it wasn't easy at the start in understanding the logic for the duplicate item checking even with spaces, but you both pulled through and done very well in smashing the AC for this! I tested all the use cases for this that included punctuation, single space, characters and symbols, and it all looks good!

src/App.tsx
```diff
@@ -70,7 +70,7 @@ export function App() {
 						/>
 						<Route
 							path="/manage-list"
-							element={<ManageList listPath={listPath} />}
+							element={<ManageList listPath={listPath} data={data || []} />}
 						/>
```
bbland1, line 74: suggestion: I don't think wee need to pass `data || []` in the data prop, I beleive the data is initialized as an empty array when creating a list. So a list would either have a filled array or an empty array.

src/utils/validateTrimmedString.ts
```diff
@@ -1,10 +1,41 @@
-// makes sure the string passed into the function isn't an empty string
-export function validateTrimmedString(input: string) {
-	const trimmedInput = input.trim();
+import { ListItem } from "../api";
 
+// Validates the item name input for a shopping list
+export function validateItemName(
```
bbland1, line 4: praise: amazing at extending functionality and bringing the new logic and the current logic together to make an awesome new piece of code!

src/utils/validateTrimmedString.ts
```diff
@@ -1,10 +1,41 @@
...
+	const trimmedInput = input.trim(); //removes leading and trailing whitespaces
+
+	// Condition 1: Check if the input is empty
 	if (trimmedInput.length === 0) {
-		return null;
+		return "Item cannot be empty";
+	}
+
+	//Remove punctuation marks and normalize input
+	const punctuationRegex = /[^\p{L}]/gu;
```
bbland1, line 16: praise: love the use of the regex, it is something I'm still such a novice to but I know it has a lot of good functionality and it is awesome to see it here!

src/utils/validateTrimmedString.ts
```diff
@@ -1,10 +1,41 @@
...
+		);
+	};
+
+	//return error if the item already exists
+	if (isDuplicateItem(normalizedInputName)) {
+		return ` ${normalizedInputName} already exists in the list`;
 	}
 
-	return trimmedInput;
+	// Return null if no errors are found (input is valid)
```
bbland1, line 39: question: is there a way without breaking everything to have this return `trimmedInput` again. I added an item to a test list that was '     hi' and it did trim it and i can't add 'hi' in any different form again I did notice in the DB it is adding the item with the white space to it instead of the `trimmedInput` like before?

bbland1: https://github.com/user-attachments/assets/6c191b26-2ab5-419f-b876-4f0bf5325282

I tried to show that the toast and such does show the trimmed value but when checking the DB there is a single whitespace before the goodbye. Sorry if it is awkward to see 😅

bbland1 (approved):

src/utils/validateTrimmedString.ts
eternalmaha: Thank you! Yeah that was definetely a major talking point when we were thinking the logic out. We thought about whether adding the duplicate item validation was expanding the specific validation function too wide, but I think it was the right call indeed :)

src/App.tsx
eternalmaha: Ahhh do you mean we wouldn't need to pass the prop at all or not need to the OR operator to suggest it could be empty?

bbland1: Sorry that was very unclear 😅 my bad I think you can have it just as `data={data}` instead of `data={data || []}`

src/utils/validateTrimmedString.ts
eternalmaha: It was really awesome to bring it back to life form the dusty attic! lol

src/utils/validateTrimmedString.ts
eternalmaha: Thats a great point. It was something we explored. I was trying to create a piece of logic with brute force, using if - else statements to return either the error if duplicate item, or return either the normalized input name. The issue was in the addItemForm, the logic is returning the object that is being returned in the validateItemName function. So it reading the normalizedinputname as an error object. Its something I definetely want to explore more. The check for the duplicate error in the addItemForm before the item is added can be refactored..

src/utils/validateTrimmedString.ts
adidalal: alternate regex option: `input.toLowerCase().replace(/[^a-z0-9]/g, '')`

adidalal (approved):

src/App.tsx
eternalmaha: Noted!! That makes sense

PR #36, opened 2024-09-20
Issue # 12: Deleting a shopping list item 

### LINKED ISSUE #12: 12. As a user, I want to be able to delete items from my shopping list so that my list isn’t cluttered with items I don’t want to buy in the future
## Summary
Users might make a mistake when entering an item, or may decide they won’t be buying a certain item again in the future. Allowing them to delete list items will help them keep their lists tidy.
 
## Acceptance criteria
- [ ] The `ListItem` component renders a button that allows the user to delete an item from their list when clicked
- [ ] Clicking the delete button prompts the user to confirm that they *really* want to delete the item
- [ ]  The `deleteItem` function in `api/firebase.js` has been filled out, and deletes the item from the Firestore database

## Notes:
* Activating the “Delete” button should show a confirmation dialog. We want to give users a chance to confirm “destructive” actions.
* It might be tempting to create your own modal dialog component to handle confirmation, but modals are fraught with accessibility concerns, so consider using a simple [JavaScript confirm dialog](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm).

eternalmaha:
## Description

Users will need to be able to delete items they no longer need to purchase or if the item was added by mistake. We need to avoid cluttering the shopping list, and only display items that fit the users' current state of needs. 

## Related Issue

closes #12 

## Acceptance Criteria

- [x] The ListItem component renders a button that allows the user to delete an item from their list when clicked
- [x] Clicking the delete button prompts the user to confirm that they really want to delete the item
- [x] The deleteItem function in api/firebase.js has been filled out, and deletes the item from the Firestore database

## Type of Changes

`new feature`
`enhancement`

## Updates

https://github.com/user-attachments/assets/988504a5-406f-4f36-9667-9bfd822541d3

1. Sign onto the application.
2. Select a list from the home view.
3. Add list items if list is empty.
4. Press the  "delete item" button to delete item.
5. A dialogue will appear to confirm deletion
6. If user clicks "yes", a confirmation message is displayed 
7. If user clicks "cancel", they are returned to the windows page as normal.

### DIFF
--- a/src/api/firebase.ts
+++ b/src/api/firebase.ts
@@ -303,7 +303,7 @@ export async function deleteItem(listPath: string, item: ListItem) {
 	// Let's try the deleteDoc from Firestore.
 	try {
 		await deleteDoc(itemDocRef);
-		alert("Item has been successfully deleted!");
+		alert(`${item.name} has been successfully deleted!`);
 	} catch (error) {
 		// If there's an error, log it to the console and throw it.
 		console.error("Oops! Error deleting Item!", error);
--- a/src/api/firebase.ts
+++ b/src/api/firebase.ts
@@ -266,10 +266,7 @@ export async function updateItem(listPath: string, item: ListItem) {
 	}
 }
 
-export async function deleteItem() {
-	/**
-	 * TODO: Fill this out so that it uses the correct Firestore function
-	 * to delete an existing item. You'll need to figure out what arguments
-	 * this function must accept!
-	 */
+//delete an item
+export async function deleteItem(listPath: string, item: ListItem) {
+	const listCollectionRef = collection(db, listPath, "items");
 }
--- a/src/api/firebase.ts
+++ b/src/api/firebase.ts
@@ -270,4 +270,14 @@ export async function updateItem(listPath: string, item: ListItem) {
 //delete an item from the list
 export async function deleteItem(listPath: string, item: ListItem) {
 	const itemDocRef = doc(db, listPath, "items", item.id);
+	// Delete the item from the list in Firestore.
+	// Should be a simple try-catch block. Try === deletes the item. Catch === logs the error.
+	// Let's try the deleteDoc from Firestore.
+	try {
+		await deleteDoc(itemDocRef);
+	} catch (error) {
+		// If there's an error, log it to the console and throw it.
+		console.error("Oops! Error deleting Item!", error);
+		throw error;
+	}
 }
--- a/src/api/firebase.ts
+++ b/src/api/firebase.ts
@@ -297,7 +297,7 @@ export async function updateItem(listPath: string, item: ListItem) {
 
 //delete an item from the list
 export async function deleteItem(listPath: string, item: ListItem) {
-	const itemDocRef = doc(db, listPath, "items", item.id);
+	const itemDocRef = doc(db, listPath, "items", item.id) || null;
 	// Delete the item from the list in Firestore.
 	// Should be a simple try-catch block. Try === deletes the item. Catch === logs the error.
 	// Let's try the deleteDoc from Firestore.
--- a/src/api/firebase.ts
+++ b/src/api/firebase.ts
@@ -297,7 +297,9 @@ export async function updateItem(listPath: string, item: ListItem) {
 
 //delete an item from the list
 export async function deleteItem(listPath: string, item: ListItem) {
+	//reference the item document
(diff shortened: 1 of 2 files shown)


src/api/firebase.ts
```diff
@@ -294,10 +295,18 @@ export async function updateItem(listPath: string, item: ListItem) {
...
-	 */
+//delete an item from the list
+export async function deleteItem(listPath: string, item: ListItem) {
+	const itemDocRef = doc(db, listPath, "items", item.id);
+	// Delete the item from the list in Firestore.
+	// Should be a simple try-catch block. Try === deletes the item. Catch === logs the error.
+	// Let's try the deleteDoc from Firestore.
+	try {
+		await deleteDoc(itemDocRef);
+		alert("Item has been successfully deleted!");
```
alex-andria, line 306: I like the alert to let the user know if an item's successfully deleted! Maybe an enhancement in the future could be to pull the item name so that the alert confirms if "Cheese has been successfully deleted!" but that is a scope creep for the acceptance criteria of this card I suppose.

alex-andria (approved):
Works as expected great work Ross and Maha! 👏 👏

src/components/ListItem.tsx
```diff
@@ -52,6 +52,20 @@ export function ListItemCheckBox({ item, listPath }: Props) {
...
 	};
 
+	const deleteItemHandler = () => {
+		const isConfirmed = window.confirm("Do you want to delete this item?");
+		console.log(isConfirmed);
+		console.log("Item:", item);
+		console.log("List Path:", listPath);
+
+		if (isConfirmed) {
+			deleteItem(listPath as string, item);
```
tannaurus, line 62: Because `deleteItem` re-throws, we'll have to catch it here and handle it too 🙂

src/api/firebase.ts
RossaMania: Fantastic suggestion! A nice touch! Simple to implement!
image

eternalmaha: Yes! We always want to make sure the user knows they are doing the action they're supposed to!

src/components/ListItem.tsx
RossaMania: Great point!
image
We wrapped the deleteItem() func call in a try-catch block! In the catch, we have the error being logged into the console! @eternalmaha pointed out we can keep notifying the user simple by doing some sort of alert! I thought we could just tuck an alert in the catch block!

eternalmaha: @RossaMania and I were just thinking about the purposes of try-catchs in our deleteItem() call. I was wrapping my head around well "if deleteItem() throws an error as part of the function", do we need to throw an error again when its called? If im understanding correctly, its because there can be an error of course when deleting the item, then there can be an error when the function is being called itself. **as I research more about try-catches and function calls ** We decided to wrap a try catch around the whole function call.

src/api/firebase.ts
```diff
@@ -266,10 +267,17 @@ export async function updateItem(listPath: string, item: ListItem) {
...
 
-export async function deleteItem() {
-	/**
-	 * TODO: Fill this out so that it uses the correct Firestore function
-	 * to delete an existing item. You'll need to figure out what arguments
-	 * this function must accept!
-	 */
+//delete an item from the list
+export async function deleteItem(listPath: string, item: ListItem) {
+	const itemDocRef = doc(db, listPath, "items", item.id);
```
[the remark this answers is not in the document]
RossaMania: Great point! We try to use nullable types ass much as possible!
image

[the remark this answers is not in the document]
RossaMania: We were able to do a little razzle dazzle like this!
image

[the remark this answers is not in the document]
eternalmaha: @RossaMania @adidalal This seems helpful :) We can play around with firebase getting documents methods 
Screenshot 2024-09-21 at 9 59 13 PM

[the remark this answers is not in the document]
eternalmaha: I see what you are saying now! The problem could be that FireBase is not getting the item to begin with. So we need to make sure that the item document is grabbed from firebase properly.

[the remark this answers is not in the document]
RossaMania: @eternalmaha brought up, since this is the file and functions talking to the firebase, we should use the .exists() method! I haven't seen that since MongoDB!
image

PR #37, opened 2024-09-22
Managing `listPath`'s impossible null state

tannaurus:
## Description

This fixes two things:
1. We were handling `listPath` being `string | null` in places where it could have been prevented already (re: making impossible states impossible). These changes take the approach of ensuring `listPath` is not `null` at the parent route level so that all sub-components don't have to contend with handling an impossible `null` state.
2. Fixes a bug where if no list was selected, the `List` page would render a component that suggested your list was just empty and redirect you to "add items to the list" which, upon being redirected, there was nothing to do (because there was no list to begin with)


src/views/authenticated/List.tsx
```diff
@@ -21,13 +21,23 @@ export function List({ data: unfilteredListItems, listPath }: Props) {
...
 
+	const Header = () => {
+		return (
+			<p>
+				Hello from the <code>/list</code> page!
+			</p>
+		);
+	};
+
+	if (!listPath) {
```
bbland1, line 32: I see that the `Header` handles the retruning early if `listPath` is null or falsy so that the components that require `listPath` to be defined don't have to do the checks themselves...

I think I am trying to wrap my head around why do that check here in the `List` & `ManageList` instead of higher in the `App`, would it have over complicated the logic or does it make conceptual sense to put it here for both of them?

bbland1 (approved):
I don't think I ever caught the bug in my testing with the list so it is awesome for it to be caught here!

I didn't really know where to put this just a random question: would us using the local storage of the `listPath` after it is set would this change how all of the passing of the `listPath` prop is handled?

src/views/authenticated/ManageList.tsx
```diff
@@ -8,11 +8,21 @@ interface Props {
 }
 
 export function ManageList({ listPath, data }: Props) {
-	return (
-		<div>
+	const Header = () => {
```
bbland1, line 11: I like this like internal component usage in creating the header. I've never really done that usually just the full pull out of a component but this is a cool thing to keep top of mind.

src/views/authenticated/List.tsx
```diff
@@ -21,13 +21,23 @@ export function List({ data: unfilteredListItems, listPath }: Props) {
...
 	// Early return if the list is empty
 	if (unfilteredListItems.length === 0) {
 		return (
 			<>
-				<p>
-					Hello from the <code>/list</code> page!
-				</p>
+				<Header />
 				<section>
 					<h2>Your list is ready!</h2>
```
eternalmaha, line 42: I like the returning of the header if no list has any items. Can we get rid of the statement "Your list is ready" if this is the case too? :) I think it might be contradicting to tell users "your list is ready" if the list is empty!

src/views/authenticated/List.tsx
tannaurus: We probably could have made those routes re-direct if `listPath` was `null`, but that would be changing existing behavior in the app. We've been displaying these routes + header if there was no `listPath`, and I didn't want to take the liberty of changing that because I think that should be the team's decision, and could be thought of as part of the project you had mentioned some interest in; there's a good amount of overlap between "what does this look like when you're not-signed in" and "what does it look like when you first sign in"

bbland1: That makes total sense, I didn't think about how we have had specific functionality and how that could change the app. Thinking about how the overall process/usage flow of the app is a consideration in where the `listPath` can be null and can't be.

tannaurus:
> I don't think I ever caught the bug in my testing with the list so it is awesome for it to be caught here!
> 
> I didn't really know where to put this just a random question: would us using the local storage of the `listPath` after it is set would this change how all of the passing of the `listPath` prop is handled?

No sweat! That's actually how I discovered this bug in the first place, I was in the process of unraveling why `listPath` could be `null` in the first place. The problem there is "after it is set", we still have to deal with "before its set" which is just unavoidable

src/views/authenticated/List.tsx
tannaurus: Done 🤝

src/views/authenticated/ManageList.tsx
tannaurus: I personally do things like this when it's a super simple thing, like a few lines that happen to repeated a few times in a given component. If it becomes any more complex, and especially if there are hooks invalid, it's much butter to move it out of the parent component. When it's this simple though, retaining the context of the parent is really nice (I don't have to name it "ManageListHeader" because it's defined inside "ManageList")

PR #44, opened 2024-09-25
Issue #15 Users should be  able to view the list name

### LINKED ISSUE #43: Issue # 43. As a user, I want to click on a list and view the name of the list I am viewing when redirected.
**Summary** 

As a user, I need to be able to view the name of the shopping list I am viewing to avoid confusion and ensure I am purchasing from the correct list. 

**Acceptance criteria**

- [ ] The user is immediately redirected to the list after clicking the name of each list. 
- [ ] The name of the list appears at the top of each list view when a user clicks.

eternalmaha:
## Description

Users should be directly taken to the list they are viewing and should be able to view the name of the list at the top view. 

## Related Issue

closes #43 (issue 15) 

## Acceptance Criteria

- [ ] The user is immediately redirected to the list after clicking the name of each list.
- [ ] The name of the list appears at the top of each list view when a user clicks.

## Type of Changes

`enhancement`

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -66,7 +66,9 @@ export function App() {
 					<Route element={<ProtectRoute user={user} redirectPath="/" />}>
 						<Route
 							path="/list"
-							element={<List data={data} listPath={listPath} />}
+							element={
+								<List data={data} listPath={listPath} listName={inputValue} />
+							}
 						/>
 						<Route
 							path="/manage-list"
--- a/src/components/CreateList.tsx
+++ b/src/components/CreateList.tsx
@@ -1,15 +1,19 @@
 import { ChangeEvent, FormEvent, useState } from "react";
-import { createList, User } from "../api";
+import { createList, ListItem, User } from "../api";
 import { useNavigate } from "react-router-dom";
+import { List } from "../views";
 import toast from "react-hot-toast";
 
 interface Props {
 	user: User;
 	setListPath: (path: string) => void;
+	listPath: string;
+	data: ListItem;
 }
 
 export function CreateList({ user, setListPath }: Props) {
 	const [inputValue, setInputValue] = useState("");
+	const [listCreated, setListCreated] = useState(false);
 	const navigate = useNavigate();
 
 	const handleSubmit = async (e: FormEvent<HTMLFormElement>) => {
@@ -27,6 +31,7 @@ export function CreateList({ user, setListPath }: Props) {
 			setListPath(path);
 			setInputValue("");
 			toast.success("Success: Your New List is Created!");
+			setListCreated(true);
 			// Delay for toast notification before redirecting
 			setTimeout(() => {
 				navigate("/list");
@@ -61,6 +66,10 @@ export function CreateList({ user, setListPath }: Props) {
 					<button aria-label="Create new shopping list">Create List</button>
 				</form>
 			</li>
+			›
+			{listCreated && (
+				<List listName={inputValue} listPath={listPath} data={data} />
+			)}
 		</>
 	);
 }
--- a/src/views/authenticated/List.tsx
+++ b/src/views/authenticated/List.tsx
@@ -1,15 +1,17 @@
 import { useState, useMemo } from "react";
 import { ListItemCheckBox } from "../../components/ListItem";
 import { FilterListInput } from "../../components/FilterListInput";
+import { CreateList } from "../../components";
 import { ListItem, comparePurchaseUrgency } from "../../api";
 import { useNavigate } from "react-router-dom";
 
 interface Props {
 	data: ListItem[];
 	listPath: string | null;
+	listName: string;
 }
 
-export function List({ data: unfilteredListItems, listPath }: Props) {
+export function List({ data: unfilteredListItems, listPath, listName }: Props) {
 	const navigate = useNavigate();
 	const [searchTerm, setSearchTerm] = useState<string>("");
 
@@ -25,9 +27,7 @@ export function List({ data: unfilteredListItems, listPath }: Props) {
 	if (unfilteredListItems.length === 0) {
 		return (
 			<>
-				<p>
-					Hello from the <code>/list</code> page!
-				</p>
+				<p>{listName}</p>
 				<section>
 					<h2>Your list is ready!</h2>
 					<h3>
@@ -49,9 +49,7 @@ export function List({ data: unfilteredListItems, listPath }: Props) {
 	// Main content when list is not empty
 	return (
 		<>
-			<p>
-				Hello from the <code>/list</code> page!
-			</p>
+			<p></p>
 
 			<div>
 				<section>


zahrafalak:
This works great! The only thing I noticed as a side effect is that the list option in the navigation bar now shows a "Page Not Found" error when clicked. The same happens when I click on `View List` button on the `ManageList` view.

src/App.tsx
```diff
@@ -65,7 +65,7 @@ export function App() {
 					{/* protected routes */}
 					<Route element={<ProtectRoute user={user} redirectPath="/" />}>
 						<Route
-							path="/list"
+							path="/list/:listName"
```
zahrafalak, line 68: One thing I noticed as a side effect is that the list option in the navigation bar now shows a "Page Not Found" error when clicked. The same happens when I click on View List button on the ManageList view. It might be happening due to this part, because we do not have a route defined for just /list anymore.

src/views/authenticated/List.tsx
```diff
@@ -1,3 +1,4 @@
+import { useParams } from "react-router-dom";
```
bbland1, line 1: On line 6 `useNavigate` is imported from `react-router-dom` again and I get an error in the terminal from vite about dupilicate imports. We should be able to combine the 2 imports to stop that error.

```suggestion
import { useParams, userNavigate } from "react-router-dom";
```

src/views/authenticated/List.tsx
```diff
@@ -25,9 +27,7 @@ export function List({ data: unfilteredListItems, listPath }: Props) {
 	if (unfilteredListItems.length === 0) {
 		return (
 			<>
-				<p>
-					Hello from the <code>/list</code> page!
-				</p>
+				<p>{listName}</p>
```
bbland1, line 30: nitpick (non-blocking): we could change this to some type of header tag. it maybe would mean adjusting the others on the page since <h1> is used in the header but it may make it a more distinct part of the page.

src/App.tsx
eternalmaha: Yeah I realized the side effect afterwards unfortunately. I'm exploring ways we can route to the list path directly in the view list navigation. I'm trying to see if I can import the object I created for list item to use it in the manage-list view. In terms of the list navigation directing to page not found, since, I'm still exploring that. Do we need the navigation to the \list navigation view anymore ? Unless there's other parts of the application that are impacted (looking into it), could we remove that navigation bar altogether since our lists are being directly accessed now through clicking them? @tannaurus @adiati98 @alex-andria

eternalmaha: I'm curious about the idea of placing all the singlelist elements into the list view altogether. Then once each list is clicked, it'll redirect to its own view?

eternalmaha: Whewww! So I was able to fix the issue of view list redirected to the 404 no page found since the \list view is out of commission for now. The name of the list is included in the end of the listPath parameter we are passing in the addItem form.  I extracted and stored the name of each list from the  listPath using the good ol - fashion pop() method! Using string interpolation, I appended our useNavigate in the addItem form  to include the name. Now we should successfully still be able to navigate to the list when clicking the "view list" button on the addItem form!

src/views/authenticated/List.tsx
eternalmaha: Yes I agree! I do not think the listName should be in a paragraph element, that's very disrespectful to it!!!

RossaMania (approved):
I like what you have done here! Even though the list name is in a paragraph element, it should be able to be styled with Bootstrap!

I do like the idea that now that the /list navigation buttons may not be needed, that it may be good to just have the /list route and buttons lead to a list of the lists. That, and the adding that ShareListForm component to each list as well. That ShareListForm component is tucked in components/forms.

As of right now, the /list nav button would no longer be needed. I wonder if all of this is scope creep and perhaps 1 or 2 new issues should be made instead?

After all, the /list route could be a list of lists and the home page could be styled up with Bootstrap and probably some sort of Call To Action thing----including a call to action to click on that list button to see the lists.

I consider this pull request approved.

src/views/authenticated/List.tsx
bbland1: I didn't realize it would put that text in as a header 😂

src/App.tsx
bbland1: if we end up wanting to keep the `/list` route we should just be able to have botht he one with the params and the basic list route as long as we order them properly. I beleive if we have the non param route first it would still work for both routes as we intend.

```
suggestion

	<Route
		path="/list"
		lement={<List data={data} listPath={listPath} />}
	/>						
    <Route
		path="/list/:listName"
		element={<List data={data} listPath={listPath} />}
	/>
```

src/views/authenticated/List.tsx
```diff
@@ -11,6 +11,7 @@ interface Props {
 
 export function List({ data: unfilteredListItems, listPath }: Props) {
 	const navigate = useNavigate();
+	const { listName } = useParams<{ listName: string }>();
```
bbland1, line 14: praise: I like seeing useParams here. It is something that I've used very little so it i nice to see it used here and getting the use that you wanted in it! 🎉

src/components/SingleList.tsx
```diff
@@ -1,13 +1,17 @@
...
-interface Props extends Pick<api.List, 'name' | 'path'> {
+interface Props extends Pick<api.List, "name" | "path"> {
 	setListPath: (path: string) => void;
 }
 
 export function SingleList({ name, path, setListPath }: Props) {
+	const navigate = useNavigate();
+
 	function handleClick() {
 		setListPath(path);
```
bbland1, line 13: suggestion: noticed that when clicking through lists when clicking one, going back to the home page and then clicking another list there is a half second moment where the old list can be seen. hopefully with the video that makes more sense. I think we could get rid of this `setListPath` since we are nvigating directly to lists with params now we don't really need to set the `listPath` for the list nav button anymore.

src/views/authenticated/List.tsx
```diff
@@ -25,9 +26,7 @@ export function List({ data: unfilteredListItems, listPath }: Props) {
 	if (unfilteredListItems.length === 0) {
 		return (
 			<>
-				<p>
-					Hello from the <code>/list</code> page!
-				</p>
+				<h1>{listName}</h1>
```
bbland1, line 29: nitpick (non-blocking): we may want to change the list name to `h2` since the header name `Smart Shopping List` is already `h1` and I guess I just had this idea that you are noly suppose to use 1 per page but as I type this I am not sure if that is a real "rule" or something I've just thought this whole time 😅🤔😂

src/components/forms/AddItemForm.tsx
```diff
@@ -85,7 +85,10 @@ export function AddItemForm({ listPath, data: unfilteredListItems }: Props) {
 		}
 	};
 	const navigateToListPage = () => {
-		navigate("/list");
+		if (listPath) {
+			const listName = listPath.split("/").pop();
```
bbland1, line 89: praise: this is a really cool way to get the `listName` from the path and a cool way to use `split` and `pop`

bbland1 (approved):

src/App.tsx
alex-andria: I'm noticing a 404 page when a new list is created:
List

src/components/SingleList.tsx
alex-andria: Yes noticed this too. I can also look more into it later tonight since this is a common routing issue.

src/views/authenticated/List.tsx
alex-andria: +1

alex-andria (commented):
Nice job on taking on this list name feature that is super helpful Maha! It works as intended now we just have to iron out a few things. When it goes to a list view, maybe some kind of `loading` spinner or logo can appear so that the user isn't seeing the half second flash of a prior list. We also need to look into the 404 error, but I can do some more digging on that tonight.

src/App.tsx
tannaurus: Towards @alex-andria 's observation, the same behavior happens when you try to navigate to the list via the `List` nav bar link. To fix that bug, we need to:
- Remove `List` from the nav bar (it no longer serves a purpose imo)
- Update `CreateList.handSubmit` to navigate to `/list:listName`

src/components/SingleList.tsx
tannaurus: The problem here is rooted in our usage of firestore's `onSnapshot` api. We're subscribing to events on the list, which allows us to view changes as they happen in real time (if I was looking at an item that I shared with Alex, and she made a change, I should see it without refreshing my list). The downside of this is we can't use Javascript Promises. We don't have a way of `await`ing the data of `useShoppingListData` to be refreshed to the new dataset.

Most importantly, I don't believe this issue was introduced in this change, it was just revealed by this change. This change makes it so much easier/quicker to jump around the app, that we're seeing how the data is just refreshing relatively slowly (we don't have a way of implementing a "loading" state).

tannaurus: Put differently, I don't think this needs to be fixed in this PR because it wasn't introduced here. Maybe someone could take a look at this issue in a follow up?

tannaurus: You'll need to do something like this

[loading_suggestion.patch](https://github.com/user-attachments/files/17175554/loading_suggestion.patch)

src/App.tsx
eternalmaha: i fixed the issue of the create list function leading to a 404 no page found. I wasn't able to navigate to /list:listName because  only because our List component route path is set to use the paramater :listName as the list name at the top of the page view and :listName appears at the top that way! I had to manipulate the "path" string to grab the name from it. 

 On the note of the "list" nav bar. It seems like the majority of the team is okay with removing the list nav bar button from the application since it no longer serves its purpose of navigating to the lists. I believe this is a separate issue, and am hesistant to remove it now unless everyone is okay with it at this point.

src/components/SingleList.tsx
eternalmaha: I really appreciate your guys' keen and thorough eyes!! I do see the lag issue. There's a brief article on firebase about the the idea of listeners.  https://firebase.google.com/docs/firestore/query-data/listen  that was a bit enlightening to read. 

I tried downloading the file and ran the git command, but It didn't work on my end. The errors on my terminal suggest there's a mismatch between the lines of code in the file and the lines on my codebase.I can try again when this turns into its own issue if need be. 

I had some question about the listpath being set in singelist not being needed. It made sense to me that we wouldn't need it if we have the params to direct us to each listview from clicking the button. Since the listPath is being supplied from our firebase, this wouldn't affect the list path  for the shareList function if we removed the setList path in singlelist right or anywhere else? I see that the path is simply being passed to singeList in the home view component as a property, singleList doesn't actually contain the data for listPath. So I believe removing listPath would be okay.

eternalmaha:
Thanks everyone for all your thoughts and insight!!! I was thinking about new issue of lag between the old list disappearing and the new list appearing. Thinking about the need to "wait" for the new dataset to refresh, I immediately thought about our good old friend the setTImeout() function. I used a small amount of time for the delay to navigate to the list so there isnt a real noticable change to view the list. This might not be the permanent solution but it seems to be a short while fix that's doing what it needs to do. 

setTimeout(() => {
			navigate(`/list/${name}`);
		}, 200)

src/views/authenticated/List.tsx
eternalmaha: I dont think thats a nitpick, I think thats valid! You mentioning this actually had me thinking about the placement of our app title "smart shopping list". We can talk about this more when we all discuss design work, but having the app name in place of the "home view" as the nav bar could be a good idea. Then the name of the list could be the central figure in the page view and take on the <h1/> tag?!

eternalmaha:
@RossaMania Thanks Ross for the feedback!  Im glad you agree that its a better user experience  for the share list component to be viewed in the same arena as the list and its contents. That would be an issue I'd be interested in exploring soon. I can open that new issue soon including the one with the list nav bar removal.

src/components/SingleList.tsx
```diff
@@ -1,13 +1,19 @@
...
+		setTimeout(() => {
```
bbland1, line 14: praise 🎉: i know it seems like there are some other ideas being discussed about how to hadnle the little life switch over delay, so if that changes 👍🏽 BUT i really want to acknowledge how smart and cool this use of the `setTimeout` is! Was just coming back to see any new chagnes and it is so smooth now!

src/components/CreateList.tsx
```diff
@@ -29,7 +29,10 @@ export function CreateList({ user, setListPath }: Props) {
 			toast.success("Success: Your New List is Created!");
 			// Delay for toast notification before redirecting
 			setTimeout(() => {
-				navigate("/list");
+				console.log(path);
```
bbland1, line 32: whose that pokemon?!

it's a...

console.log!!

😂😅

src/App.tsx
bbland1: @eternalmaha I'm personally gonna vote/say go ahead and remove it. I feel like this PR really progressed the functionallity of the list name button and made the nav bar list a bit redundant and I think it would be ok to do here as well.

eternalmaha: removed!!

src/views/authenticated/List.tsx
eternalmaha: I was very excited to learn about its existance when I was exploring routing concepts that were more flexible! Thanks for the appreciation!

src/components/forms/AddItemForm.tsx
eternalmaha: shout out to my  past life of freeCodeCamp  tutorial rabbit holes!

src/components/CreateList.tsx
eternalmaha: LOL! I guess I didn't catch em all !!!

PR #46, opened 2024-09-27
Feat/add item quantity

RossaMania:

## Description

This code adds "item quantity functionality"!
- I add item quantity parameter to addItem function in firebase.ts. It's been updated to include an additional parameter, itemQuantity. This allows for specifying the quantity of the item being added to the list. I figure this would be as close as I would get to updating some sort of controller function.
- I add item quantity parameter to addItem function in AddItemForm.tsx in the AddItemForm.tsx file on what would be the frontend. Add itemQuantity and setItemQuantity state var and state func with the useState hook. Console log item quanity of 1 upon successful addition. Set item quantity state to 1 upon success. Update return string interpolation in toast to show itemQuantity as well. When an item is added successfully, the default quantity should be “1” in the browser console log and in the toast notification. Upon completion, the state will still be set to “1”.
- The quantity successfully shows up in the browser console and the toast! The itemQuantity shows up in the Firebase itself with the item! I add quantity input to AddItemForm.tsx with type "number". I wired the input up to handleItemQuantityChange func. I add and revise console logs for item quantity entered in that input to make sure the value is captured, and item quantity added to the list.
- I add the item quantity property to ListItemModel in firebase.ts. I learned this must be the actual data model in Firebase! The itemQuantity property represents the quantity of the item. It is of type "number". Once the data model has that itemQuantity, I am able to render item.itemQuantity into the ListItem.tsx view!
- This feature is getting pretty big already! I add item quantity input to AddItemForm.tsx and create ItemQuantityForm component. I add a quantity input field to the AddItemForm.tsx component with the type "number". I create a new component called ItemQuantityForm.tsx to handle the quantity input. The ItemQuantityForm component includes a form with an input field for the quantity. The quantity input field has a maximum value of 100, is initially set to “1”, can be edited and saved, and the saved quantity is passed to the parent component using a callback function. The edit is a piece of local state set with the useState hook. When edit is disabled, the input is grayed out. I also add console logs to track the quantity entered and the quantity saved.
- I rendered the ItemQuantityForm component in the ItemList.tsx file!
- I created a updateItemQuantity func in firebase.ts! This takes a quantity argument in addition to the item and listPath arguments! It's called in the ListItem.tsx!
- "item" is passed as a prop to the ItemQuantityForm from the AddItemForm.tsx file and the ListItem.tsx file so the item.itemQuantity prop can display the quantity dynamically!
- In ListItem.tsx, a check is set so if the  quantity is set to "0", the deleteItemHandler function is called as if the user clicked the delete button and the user is prompted to confirm if he or she really wants to delete the item! 

## Acceptance Criteria

- [x] "number" is going to be the data type.
- [x] "1" should be the default quantity.
- [x] The quantity should be saved to the database.
- [x]  Setting the quantity to "0" prompts the user to confirm that they want to delete the item from the list.

## Type of Changes

### Before
image

### After
image

image

image

image

image

- Plenty of console logs right now to look at that delicious data flow!

### Change Quantity!
- Navigate to the Manage List page.
- Click that Edit button. Enter a number in "How Many" text field. Click Save. 
- Enter an item name, and select an option for when to buy.
- Then click Submit.
- On the /list page, you can click that Edit button, enter a new number, and then click Save!

### 1 Is The Default Number!
- Navigate to the Manage List page.
- Leave the "How Many" text field blank!
- Enter an item name, and select an option for when to buy.
- Then click Submit.
- You should see a confirmation toast that only one of your item has been added!

### Delete Item!
- Navigate to the List page of a list with items that have quantities.
- Click that Edit button. Enter "0" in "How Many" text field. Click Save. 
- The deleteItemHandler func will be called, and you can go through that flow like normal!
- A dialogue will appear to confirm deletion
- If user clicks "yes", a confirmation message is displayed.
- If user clicks "cancel", they are returned to the windows page as normal.


src/api/firebase.ts
```diff
@@ -95,6 +95,7 @@ export function useShoppingLists(user: User | null) {
 const ListItemModel = t.type({
 	id: t.string,
 	name: t.string,
+	itemQuantity: t.number,
```
bbland1, line 98: praise: nice use of expanding the exiting model!

src/components/forms/ItemQuantityForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+					aria-label="Item quantity"
+					type="number"
+					name="item-quantity"
+					max="100"
+					value={itemQuantity}
+					onChange={(e) => setItemQuantity(Number(e.target.value))}
+					disabled={!edit}
+				/>
+				<div>
+					{edit ? (
```
bbland1, line 49: suggestion: removing these divs allows the edit button to be inline with the input which does make it a bit easier for me to understand what is being edited in general but mostly on the list page itself.

src/components/forms/ItemQuantityForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+		e.preventDefault();
+		setEdit(!edit);
+		saveItemQuantity(itemQuantity);
+		console.log("Item quantity saved:", itemQuantity);
+	};
+
+	return (
+		<section>
+			<form>
+				<label htmlFor="item-quantity">How many:</label>{" "}
```
bbland1, line 37: question: what do we think of "quantity" for the label?

src/components/ListItem.tsx
```diff
@@ -83,6 +84,30 @@ export function ListItemCheckBox({ item, listPath }: Props) {
 		}
 	};
```
bbland1, line 86: suggestion: i think we could move this logic for updating the item quantity into the `ItemQuantityForm`. reading through the code right now it feels liek tehre is some duplication of logic, kind of adding logic to components to make another component work. 

moving this to be handled by the `ItemQuantityForm` it could make it a bit more reusable because we could just place it into the location it is needed and not write logic within that component to get it to work. it owuld be like this check box and other forms the logic to do the updating and steps the form is doing is housed within the logic.

To get the `item` prop needed with these changes on the list view we could try moving the `<ItemQuantityForm saveItemQuantity={editItemQuantity} item={item} /> x{" "}` call into the `List` component with this logic ```{filteredListItems.map((item) => (<ListItemCheckBox key={item.id} item={item} listPath={listPath} />))}``` and I believe it would get the same result.

src/components/forms/ItemQuantityForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+	saveItemQuantity: (quantity: number) => void;
+	item: ListItem;
+}
+
+const ItemQuantityForm: React.FC<ItemQuantityFormProps> = ({
+	saveItemQuantity,
+	item,
+}) => {
+	// A state variable to store the item quantity.
+	const [itemQuantity, setItemQuantity] = useState<number>(item?.itemQuantity);
```
bbland1, line 14: suggestion: since the default quantity is 1 for all items it may be nice to have this value be set to that so on the adding item form it shows what it will be for the user.

```suggestion
const [itemQuantity, setItemQuantity] = useState<number>(item ? item.itemQuantity : 1);
```

bbland1:
I wasn't sure where or how to put this as a comment on a file, but when running the code with `tsc --noEmit --watch` I was getting this error in the terminal.

Screenshot 2024-09-28 at 4 46 16 AM

src/components/forms/AddItemForm.tsx
```diff
@@ -63,23 +76,30 @@ export function AddItemForm({ listPath, data: unfilteredListItems }: Props) {
 			return;
 		}
 
+		if (itemQuantity < 1) {
```
bbland1, line 79: praise: nice catch for if someone does edit the quantity when adding and messes up!

src/api/firebase.ts
tannaurus: Yes! io-ts looks very unfriendly at first put with a bit of practice it becomes less-so! I highly recommend anyone here take a look into the documentation. Feel free to shoot me any questions about it 😃  getting comfortable with it will help a ton

src/components/ListItem.tsx
```diff
@@ -83,6 +84,30 @@ export function ListItemCheckBox({ item, listPath }: Props) {
...
 
+	const editItemQuantity = async (quantity: number) => {
+		console.log("Item quantity edited:", quantity);
+
+		if (!listPath) {
```
tannaurus, line 90: do we still need to be checking this after my PR merged? I wonder if I missed something 🤔

src/components/forms/ItemQuantityForm.tsx
tannaurus: Great suggestion! I'm a little curious why item could be undefined in the first place 👀  do we need the question mark at all?

src/components/ListItem.tsx
```diff
@@ -83,6 +84,30 @@ export function ListItemCheckBox({ item, listPath }: Props) {
...
+			toast.error("Error: listPath is missing or invalid.");
+			return;
+		}
+
+		if (quantity === 0) {
+			deleteItemHandler();
+		}
```
alex-andria, line 97: Hmm I'm thinking it may be a smoother user experience if we don't give them the option to put 0 and just regularly delete the item with the delete button. Let me know if folks think otherwise, but I do appreciate the thoughtfulness of it.

src/components/ListItem.tsx
alex-andria: Great idea, that way the logic outside of firebase for quantity is all referable in one place.

src/components/forms/ItemQuantityForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+	// A function that will save the item quantity.
+	const updateItemQuantity = (e: FormEvent<HTMLButtonElement>): void => {
+		e.preventDefault();
+		setEdit(!edit);
+		saveItemQuantity(itemQuantity);
+		console.log("Item quantity saved:", itemQuantity);
+	};
+
+	return (
+		<section>
```
tannaurus, line 35: we might not need this `section` 👀

src/components/forms/AddItemForm.tsx
```diff
@@ -63,23 +76,30 @@ export function AddItemForm({ listPath, data: unfilteredListItems }: Props) {
...
+			toast.error("Oops! Item quantity must be more than 0!");
+			return;
+		}
```
alex-andria, line 82: Found a bug:
After submitting one item with a specific quantity, staying on the same page to submit a new item I am unable to edit the quantity to a different number. I also see the `0` quantity get frozen there. See my replication below:
Sep-28-2024 10-43-05

src/components/forms/ItemQuantityForm.tsx
alex-andria: +1

src/components/forms/ItemQuantityForm.tsx
alex-andria: This makes sense if the user already answered the `how many` prompt in the add item view. Also "Amount" could work too. Up to you guys!

alex-andria (changes requested):
This is a cool feature Ross, I haven't seen a Smart Shopping List from TCL with a quantity feature so it's awesome to see you add this functionality and be so motivated to do so. I found a bug when adding a second item with a different quantity that needs to be addressed, and maybe some refactoring like Brianna suggested could help with reading this as well on a dev level. Great work with the amount of detail you added in your PR too!

src/components/forms/ItemQuantityForm.tsx
```diff
@@ -0,0 +1,60 @@
+import { FormEvent, useState } from "react";
+import { ListItem } from "../../api";
+
+interface ItemQuantityFormProps {
+	saveItemQuantity: (quantity: number) => void;
+	item: ListItem;
+}
+
+const ItemQuantityForm: React.FC<ItemQuantityFormProps> = ({
```
tannaurus, line 9: minor nit pick: all of our other comments look like they're not exporting "default" -> they're exporting the components with this pattern: `export function ItemQuantityForm`

src/components/forms/ItemQuantityForm.tsx
tannaurus: I like that!

src/api/firebase.ts
RossaMania: Thank you for noticing! For a few minutes there, I was all, "Is this the data model? This has got to be the data model schema thing!"

src/components/forms/ItemQuantityForm.tsx
RossaMania: Thank you for noticing! I agree! 

There are too many divs, and I was actually going to spend some time today to figure out which divs I wanted to remove! I have removed these!

image

src/components/ListItem.tsx
RossaMania: Thank you for noticing!

I was actually wondering the same thing!!!!

I only have it in there because other funcs had it in there (for adding an item, for example) but yeah. I was kind of wondering that. I left it in to be better safe than sorry! I will make sure my branches are up to date,  remove it, and make sure the code still runs!

Oh man! Doing all of this resulted in getting extra practice with Git!

After resolving some merge conflicts, code still runs! 

image

src/components/forms/ItemQuantityForm.tsx
RossaMania: Thank you for noticing!

I couldn't decide if it should be wrapped in a ```section``` or ```form```! I will remove the extra ```section``` element!

RossaMania: ```section``` removed!
image

src/components/forms/ItemQuantityForm.tsx
RossaMania: Thank you for noticing!
Sure! I just wanted the label to be as specific as possible! 
I'm happy to change it! I will just rename it to "quantity"!
image
(.................................."amount" makes me think of like a money amount, rather than a list or cart quantity)

src/components/forms/ItemQuantityForm.tsx
RossaMania: Thank you for noticing, and letting me know!!!!

It's not a nitpick!!! I thought I fixed this before I committed and created the pull request!!!

Yes! Consistency is good! We can't have a default export in a forest of named exports! 

image

src/components/forms/ItemQuantityForm.tsx
RossaMania: I needed the "?" to make it a nullable type because if there was no items on the list, an error would be thrown in the browser about "item" being null or undefined.

.................Because I was overthinking a logical operator for useState<number>(item.itemQuantity || 1) . 🤣 🤣 🤣 🤣 🤣

RossaMania: > suggestion: since the default quantity is 1 for all items it may be nice to have this value be set to that so on the adding item form it shows what it will be for the user.

Yes please! Thank you for noticing! I was trying to do that with some logical operators. I should have just tried out ternary operators!

RossaMania: > +1

💯

src/components/forms/AddItemForm.tsx
RossaMania: Thank you for noticing! It's always nice to be able to put in "guards" or "error boundaries" or "authentication"--whatever someone would call it!

src/components/ListItem.tsx
RossaMania: Thank you for noticing, and thank you for the feedback!

WIP: I removed the deleteItemHandler call when the quantity is equal to "0". I added a "min" attribute to ItemQuantityForm's input and set it to 1.

image
image

RossaMania: I basically went back to my original premise: I want to keep a "separation of concerns". Sure, re-usable components are cool too. 

SO
I made the item quantity just another input on the Add Item Form.
image
This resolved A LOT of bugs, including features that weren't happening that were supposed to be happening with the attribute being set to "number". 
I also took the time to remove the "How many?" label from the input of the InputQualityForm component.
image

src/components/forms/AddItemForm.tsx
RossaMania: Thank you for noticing! And testing!
All of this should be resolved now that I set the item quantity as it's own label and input in the Add Item form. The number input is working properly, which is answering ALL of my questions, and fixing most of the bugs/stuff I was worried about.

Plus, when the form is submitted, all the inputs are reset.

...........I thought my replies and screenshots were going to this.

RossaMania:
I was able to merge other pull requests into this pull request as well, and refactor as needed!
image
image
image

Still a bit of refactoring to do!

RossaMania:
Ope! I forgot to put in a screenshot of what the /list screen looks like now that the item.name interpolation is removed from the label element.

image

src/components/ListItem.tsx
```diff
@@ -80,6 +81,26 @@ export function ListItemCheckBox({ item, listPath }: Props) {
 		}
 	};
 
+	const editItemQuantity = async (quantity: number) => {
```
bbland1, line 84: More a comment for later if we ever do a refactor. It may still be beneficial to move the logic for updating the item quantity to be contained in the `ItemQuanitiyForm`

Nonblocking!

Awesome functionality to add to the app!

bbland1 (approved):

